PR #5, opened 2016-04-13
move to scopes

stephanenicolas:
Includes #4 

Migration to scope concept, much more clear. 
Demonstrates the invariance of injectors and makes the API more ethereal even.
Syntactic sugar added... in ToothPick, not smoothie :)

PR #7, opened 2016-04-13
plugin named

stephanenicolas:
Named works. Built on top of both #4 and #5.


dlemures:
👍 merge after solving conflicts.
You can merge this PR directly instead of the previous 2

PR #9, opened 2016-04-14
Jake: When you specify a license it only accepts those, when you spec…

stephanenicolas:
…ify no license it accepts them all (or so I think).


dlemures:
👍

dlemures:
merge when green 😄

PR #15, opened 2016-04-16
Wrong detection of multiple annotated constructors.

stephanenicolas:
Open PR from migration of MGA to TP, should always be mergeable but might have some work in progress


dlemures:
👍

PR #16, opened 2016-04-17
bug fixes from migration of MGA to TP

stephanenicolas:
second batch. We now clear the naming system at large as we had some tech debt here. API looks much nicer and we could kill the usage of java.lang.Object.


dlemures:
👍

PR #18, opened 2016-04-17
Sni/mga tp bug fix 4

stephanenicolas:
Refactor the compiler.

The idea is that we should allow instances, providers and lazy to be available for fields, methods params and constructor params like : 
```java
class A {
  @Inject Foo f;
  @Inject Provider<Foo> f;
  @Inject Lazy<Foo> l;

  @Inject
  public A(Foo f, Provider<Foo> p, Lazy<Foo> l) {...}

  @Inject
  public void m(Foo f, Provider<Foo> p, Lazy<Foo> l) {...}
}
```

It implied a few things on the structure of the generators, this PR cleans up all code after refactor.


dlemures:
👍

PR #21, opened 2016-04-18
Some small fixes and style

dlemures:
Some small fixes and style

### DIFF
--- a/toothpick-runtime/src/main/java/toothpick/ToothPick.java
+++ b/toothpick-runtime/src/main/java/toothpick/ToothPick.java
@@ -1,6 +1,7 @@
 package toothpick;
 
-import java.util.concurrent.ConcurrentHashMap;
+import java.util.HashMap;
+import java.util.Map;
 
 /**
  * Main class to access toothpick features.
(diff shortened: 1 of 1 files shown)


toothpick-runtime/src/main/java/toothpick/ToothPick.java
```diff
@@ -36,41 +36,46 @@ public static Scope openScopes(Object... names) {
...
   public static void reset() {
-    MAP_KEY_TO_INJECTOR.clear();
+    MAP_KEY_TO_SCOPE.clear();
   }
 
   public static void inject(Object obj, Scope scope) {
     injector.inject(obj, scope);
   }
+
+  private static void removeScopeAndChildrenFromMap(Scope scope) {
```
stephanenicolas: you should add a comment to say this method is not synchronized.

toothpick-runtime/src/main/java/toothpick/ToothPick.java
```diff
@@ -36,41 +36,46 @@ public static Scope openScopes(Object... names) {
...
+      if (scope == null) {
+        scope = new ScopeImpl(name);
+        MAP_KEY_TO_SCOPE.put(name, scope);
       }
+      return scope;
     }
-    return scope;
   }
 
-  public static void closeScope(Object key) {
```
stephanenicolas: ok

toothpick-runtime/src/main/java/toothpick/ToothPick.java
dlemures: added

dlemures:
ConcurrentHashMap not needed, we need to synchronize where the map is used anyway.

There was the option to keep the ConcurrentHashMap to allow multiple "closeScope" at the same time for some special cases. However, it is too complicated and asymmetric (open vs close), so the improvement it is not worth.

PR #22, opened 2016-04-18
Sni/mga tp bug fix 6

stephanenicolas:
There was a bug in the optimistic factory creation. We had to match more the Guice style and provider factories for classes with a constructor by default, and possibly others, but the factory will use the default one.


dlemures:
👍

PR #26, opened 2016-04-19
Enhance optimistic factory creations. Classes containing @Inject anno…

stephanenicolas:
…tated members will also have factories. Code needs clean up

Not sure I want to merge it right now, code is not clean yet.


dlemures:
👍

PR #28, opened 2016-04-20
Sni/mga tp bug fix open mga pr5

stephanenicolas:
Many synchronization issues solved, lots of docs added and rationalized.


dlemures:
👍

PR #29, opened 2016-04-22
Fixing binding to allow providers from child classes

dlemures:
Fixing binding to allow providers from child classes


stephanenicolas:
RTM

PR #45, opened 2016-05-01
Sni/lock free

stephanenicolas:
DO NOT MERGE

Besides Daniel, I would love Michael & Henri to review this.
The goal of this PR is to 
* make TP crash free in a multi threading context. It appears that we have been making a few mistakes on this since the beginning. 
* pass tests that heavily use the data structure (scope tree) concurrently
* make TP lock free. Each injection will see a snapshot of TP scope tree, before or after any modification but not in between. This makes the most normal usages of TP much faster.

This version provides an improvement for massive DI  : 

image

For 55 K injections.

### DIFF
--- a/toothpick-runtime/src/main/java/toothpick/ScopeImpl.java
+++ b/toothpick-runtime/src/main/java/toothpick/ScopeImpl.java
@@ -29,6 +29,8 @@ import static java.lang.String.format;
  * </p>
  */
 public class ScopeImpl extends Scope {
+  private static final String LINE_SEPARATOR = System.getProperty("line.separator");;
+
   private static IdentityHashMap<Class, UnNamedAndNamedProviders> mapClassesToUnBoundProviders = new IdentityHashMap<>();
   protected IdentityHashMap<Class, UnNamedAndNamedProviders> mapClassesToAllProviders = new IdentityHashMap<>();
   private boolean hasTestModules;
@@ -80,13 +82,13 @@ public class ScopeImpl extends Scope {
   }
 
   @Override
-  public void installTestModules(Module... modules) {
-    //we allow multiple calls to this method
-    boolean oldHasTestModules = hasTestModules;
+  public synchronized void installTestModules(Module... modules) {
+    if (hasTestModules) {
+      throw new IllegalStateException("TestModules can only be installed once per scope.");
+    }
     hasTestModules = false;
     installModules(modules);
-    boolean doOverrideModulesExist = modules != null;
-    hasTestModules = oldHasTestModules || doOverrideModulesExist;
+    hasTestModules = true;
   }
 
   @Override
@@ -107,52 +109,61 @@ public class ScopeImpl extends Scope {
     builder.append(name);
     builder.append(':');
     builder.append(System.identityHashCode(this));
-    builder.append('\n');
+    builder.append(LINE_SEPARATOR);
 
     builder.append("Providers: [");
-    ArrayList<Class> scopedProviders = new ArrayList(mapClassesToAllProviders.keySet());
-    Collections.sort(scopedProviders, new ClassNameComparator());
-    for (Class aClass : scopedProviders) {
+    ArrayList<Class> sortedBoundProviderClassesList;
+    synchronized (mapClassesToAllProviders) {
+      sortedBoundProviderClassesList = new ArrayList(mapClassesToAllProviders.keySet());
+      Collections.sort(sortedBoundProviderClassesList, new ClassNameComparator());
+    }
+    for (Class aClass : sortedBoundProviderClassesList) {
       builder.append(aClass.getName());
       builder.append(',');
     }
-    if (!mapClassesToAllProviders.isEmpty()) {
+
+    if (!sortedBoundProviderClassesList.isEmpty()) {
       builder.deleteCharAt(builder.length() - 1);
     }
-    builder.append(']');
-    builder.append('\n');
 
-    Iterator<Scope> iterator = childrenScopes.iterator();
+    builder.append(']');
+    builder.append(LINE_SEPARATOR);
+
+    Iterator<Scope> iterator = childrenScopes.values().iterator();
     while (iterator.hasNext()) {
       Scope scope = iterator.next();
       boolean isLast = !iterator.hasNext();
       builder.append(isLast ? lastNode : node);
       builder.append(branch);
       String childString = scope.toString();
-      String[] split = childString.split("\n");
+      String[] split = childString.split(LINE_SEPARATOR);
       for (int i = 0; i < split.length; i++) {
         String childLine = split[i];
         if (i != 0) {
           builder.append(indent);
         }
         builder.append(childLine);
-        builder.append('\n');
+        builder.append(LINE_SEPARATOR);
       }
     }
 
     if (getRootScope() == this) {
-      builder.append("UnScoped providers: [");
-      ArrayList<Class> unscopedProviders = new ArrayList(mapClassesToUnBoundProviders.keySet());
-      Collections.sort(unscopedProviders, new ClassNameComparator());
-      for (Class aClass : unscopedProviders) {
+      builder.append("Unbound providers: [");
+      ArrayList<Class> sortedUnboundProviderClassesList;
+      synchronized (mapClassesToUnBoundProviders) {
+        sortedUnboundProviderClassesList = new ArrayList(mapClassesToUnBoundProviders.keySet());
+        Collections.sort(sortedUnboundProviderClassesList, new ClassNameComparator());
+      }
+
+      for (Class aClass : sortedUnboundProviderClassesList) {
         builder.append(aClass.getName());
         builder.append(',');
       }
-      if (!mapClassesToUnBoundProviders.isEmpty()) {
+      if (!sortedUnboundProviderClassesList.isEmpty()) {
         builder.deleteCharAt(builder.length() - 1);
       }
       builder.append(']');
-      builder.append('\n');
+      builder.append(LINE_SEPARATOR);
     }
 
     return builder.toString();
@@ -161,19 +172,17 @@ public class ScopeImpl extends Scope {
   private void installModule(Module module) {
     for (Binding binding : module.getBindingSet()) {
       if (binding == null) {
-        throw new IllegalStateException("A module can't have a null binding.");
+        throw new IllegalStateException("A module can't have a null binding : " + module);
       }
 
       Class clazz = binding.getKey();
-      synchronized (clazz) {
-        String bindingName = binding.getName();
-        if (!hasTestModules || getBoundProvider(clazz, bindingName) == null) {
-          InternalProviderImpl provider = toProvider(binding);
-          if (binding.isScoped()) {
-            installScopedProvider(clazz, bindingName, (ScopedProviderImpl) provider);
-          } else {
-            installBoundProvider(clazz, bindingName, provider);
-          }
+      String bindingName = binding.getName();
+      if (!hasTestModules || getBoundProvider(clazz, bindingName) == null) {
+        InternalProviderImpl provider = toProvider(binding);
+        if (binding.isScoped()) {
+          installScopedProvider(clazz, bindingName, (ScopedProviderImpl) provider);
+        } else {
+          installBoundProvider(clazz, bindingName, provider);
         }
       }
     }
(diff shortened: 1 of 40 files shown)


toothpick-runtime/src/main/java/toothpick/ToothPick.java
```diff
@@ -68,26 +76,25 @@ public static Scope openScope(Object name) {
...
-    synchronized (ToothPick.class) {
-      Scope scope = MAP_KEY_TO_SCOPE.get(name);
-      if (scope != null) {
-        Scope parentScope = scope.getParentScope();
-        if (parentScope != null) {
-          parentScope.removeChild(scope);
-        }
-        removeScopeAndChildrenFromMap(scope);
+
+  public static synchronized void closeScope(Object name) {
```
mmadev: do you still need `synchronized` here? or only synchronized when `scope != null`

toothpick-runtime/src/main/java/toothpick/UnScopedProviderImpl.java
```diff
@@ -44,60 +46,65 @@ public UnScopedProviderImpl(Class<?> factoryKeyClass, boolean isProviderFactoryC
...
-      return instance;
-    }
-
-    if (providerInstance != null) {
-      if (isLazy) {
-        instance = providerInstance.get();
-        //gc
-        providerInstance = null;
+  public T get(Scope scope) {
+    lockGet.lock();
```
mmadev: can you explain? I don't see any benefit of using ReentrantLock instead of synchronized

toothpick-runtime/src/main/java/toothpick/util/ConcurrentIdentityHashMap.java
```diff
@@ -0,0 +1,1395 @@
...
+import java.util.concurrent.locks.ReentrantLock;
+
+/**
+ * An alternative identity-comparing {@link ConcurrentMap} which is similar to
+ * {@link ConcurrentHashMap}.
+ *
+ * @param <K> the type of keys maintained by this map
+ * @param <V> the type of mapped values
+ */
+public final class ConcurrentIdentityHashMap<K, V> extends AbstractMap<K, V> implements ConcurrentMap<K, V> {
```
mmadev: wow!!! is this the ConcurrentHashMap source code? what's the difference between the ConcurrentHashMap and your version?

toothpick-runtime/src/main/java/toothpick/ToothPick.java
Macarse: why `openScope()` is not synchronized and `closeScope()` is?

toothpick-runtime/src/main/java/toothpick/ToothPick.java
```diff
@@ -68,26 +76,25 @@ public static Scope openScope(Object name) {
...
       }
+      removeScopeAndChildrenFromMap(scope);
     }
   }
 
   /**
    * Clears all scopes. Useful for testing and not getting any leak...
    */
+
   public static void reset() {
```
Macarse: Why don't you change this to package visibility so nobody uses it.

toothpick-runtime/src/main/java/toothpick/ToothPick.java
```diff
@@ -100,11 +107,27 @@ public static void inject(Object obj, Scope scope) {
     injector.inject(obj, scope);
   }
 
-  // Not synchronized, called by closeScope that is synchronized
```
Macarse: You should leave this comment or add something like:
This method is not thread safe.

toothpick-runtime/src/main/java/toothpick/ToothPick.java
```diff
@@ -100,11 +107,27 @@ public static void inject(Object obj, Scope scope) {
...
+   * Removes all nodes of {@code scope} using DFS. We don't lock here.
+   *
+   * @param scope the parent scope of which all children will recursively be removed
+   * from the map. We don't do anything else to the children nodes are they will be
+   * garbage collected soon. We just cut a whole sub-graph in the references graph of the JVM normally.
+   */
   private static void removeScopeAndChildrenFromMap(Scope scope) {
     MAP_KEY_TO_SCOPE.remove(scope.getName());
-    for (Scope childScope : scope.childrenScopes) {
+    for (Scope childScope : scope.childrenScopes.values()) {
```
Macarse: isn't there a way of going through the map values without the need of synchronized?

toothpick-runtime/src/main/java/toothpick/ToothPick.java
```diff
@@ -100,11 +107,27 @@ public static void inject(Object obj, Scope scope) {
...
     }
   }
+
+  /**
+   * This method is for testing only.
+   * It is not thread safe and WILL CRASH in multi-thread.
+   * TODO how can we hide this API ? Reflection in tests ?
+   */
+  @Deprecated
+  public static int getScopeNamesSize() {
```
Macarse: package scope.

toothpick-runtime/src/main/java/toothpick/ToothPick.java
```diff
@@ -52,14 +57,17 @@ public static Scope openScopes(Object... names) {
...
    */
   public static Scope openScope(Object name) {
-    synchronized (ToothPick.class) {
-      Scope scope = MAP_KEY_TO_SCOPE.get(name);
-      if (scope == null) {
-        scope = new ScopeImpl(name);
-        MAP_KEY_TO_SCOPE.put(name, scope);
-      }
+    Scope scope = MAP_KEY_TO_SCOPE.get(name);
+    if (scope != null) {
```
dlemures: We have:
... A
../ ...\
B.....C

And then:
1. Thread 1 calls -> closeScope(A)
2. Thread 2 calls -> openScope(C)

Thread 2 can get scope C when it is about to be removed. Should we allow that?

toothpick-runtime/src/main/java/toothpick/ToothPick.java
```diff
@@ -100,11 +107,22 @@ public static void inject(Object obj, Scope scope) {
...
-  // Not synchronized, called by closeScope that is synchronized
+  /**
+   * Removes all nodes of {@code scope} using DFS. We don't lock here.
+   *
+   * @param scope the parent scope of which all children will recursively be removed
+   * from the map. We don't do anything else to the children nodes are they will be
+   * garbage collected soon. We just cut a whole sub-graph in the references graph of the JVM normally.
+   */
   private static void removeScopeAndChildrenFromMap(Scope scope) {
     MAP_KEY_TO_SCOPE.remove(scope.getName());
```
dlemures: It can be moved inside the for now

toothpick-runtime/src/main/java/toothpick/UnScopedProviderImpl.java
stephanenicolas: They are just faster. But I agree, it's a very minor change here.
http://lycog.com/concurency/performance-reentrantlock-synchronized/

toothpick-runtime/src/main/java/toothpick/util/ConcurrentIdentityHashMap.java
stephanenicolas: this one is an identity hash map. This code had to be copy-pasted from netty. 
An identity map uses == on keys rather than equals basically, which is faster and more suited to a map with key of type java.lang.Class.

toothpick-runtime/src/main/java/toothpick/ToothPick.java
stephanenicolas: it's needed for tests. That was one of the reasons of the latest leaks in RG during testing.

toothpick-runtime/src/main/java/toothpick/ToothPick.java
stephanenicolas: yes, it's a concurrent data structure. The iterators are fail safe, not fail fast. They are basically returning a snapshot of the iterables.

toothpick-runtime/src/main/java/toothpick/UnScopedProviderImpl.java
mmadev: there is bug is the article, so the test result may not true.
```
While I was reading your code I’ve realized the way you were using the ReentrantLock was wrong, because you were creating a new lock every time. If you are not using the same lock for each thread then it is similar to not having a lock. There is another comment pointing out this, but the bug is still there. It would be better if you could fix it, otherwise people might get confused if they do not read the comments.
```

stephanenicolas: Ok, but still the benchmark has shown a good improvement in perfs, probably not due to this single change.. But as stated in this article, in the end, http://stackoverflow.com/a/11821900/693752
there is also a possibility to go even further and use read and write locks in a future. And I prefer having all the code using locks all the time than a mix of locks and synchronized. 

But you are right, my first argument was wrong, it looks like it has never been truly benchmarked ???

stephanenicolas: http://blog.takipi.com/java-8-stampedlocks-vs-readwritelocks-and-synchronized/

this one says synchronized is better for a low number of threads, which will be the case in Android. LOL. No one agrees :) 
I would keep the lock until someone else comes with a proper synchronized approach and benchmarks to say it's better... Apparently we got an improvement in speed using locks and the new lock free DS is actually well synchronized I believe. So it cannot really be compared to what we used to have which was not thread safe, or not totally.

toothpick-runtime/src/main/java/toothpick/ToothPick.java
stephanenicolas: yes, t2 should get something. The idea of the lock free structures is that when a thread reads and another thread  writes, the reader sees the data structure either before or after the change, but not during the transition. And I believe it's the case here. In a multi-threaded context, you should be careful that your scope was not closed before using it. 

Closing a scope will detach it from the tree, and C will become a root of a new tree indeed. With synchronized, the exact same problem will happen a millisecond later if some thread closes the scope you just opened.

My point of view is that this can be more or less adressed by DI itself. In Android for instance, the main thread starts a background threads with a callback, when the main thread leaves the context, it should remove the callback and interrupt the background thread or not. If the background thread uses only the application level scope, it can be allowed to continue will the activity is dead, but if it uses activity scope or lower, it should be stopped. 

So in one case, you can use the scope, in the other not. Really, I don't think this can be fully solved by DI. What I favored here is performance.

stephanenicolas:
Only coveralls fails (-0.4%) but build passes.

toothpick-runtime/src/main/java/toothpick/ScopeImpl.java
```diff
@@ -436,9 +436,8 @@ private void installModule(Module module) {
       return getProvider(UN_NAMED_PROVIDER_KEY);
     }
 
-    public synchronized InternalProviderImpl<? extends T> getProvider(String bindingName) {
-      InternalProviderImpl<? extends T> internalProvider = mapNameToProvider.get(bindingName);
-      return internalProvider;
+    public InternalProviderImpl<? extends T> getProvider(String bindingName) {
```
mmadev: your implementation may not crash since it only write once. But it may crash or return unexpected value when the hashmap is resizing
http://grepcode.com/file/repository.grepcode.com/java/root/jdk/openjdk/6-b14/java/util/HashMap.java#390
http://grepcode.com/file/repository.grepcode.com/java/root/jdk/openjdk/6-b14/java/util/HashMap.java#HashMap.resize%28int%29

toothpick-runtime/src/main/java/toothpick/ScopeImpl.java
```diff
@@ -390,47 +389,66 @@ private void installModule(Module module) {
...
-        if (mapNameToProvider == null) {
-          mapNameToProvider = new HashMap<>();
-          unNamedAndNamedProviders.setMapNameToProvider(mapNameToProvider);
+    UnNamedAndNamedProviders<T> unNamedAndNamedProviders = map.get(clazz);
+    if (unNamedAndNamedProviders == null) {
+      synchronized (map) {
+        unNamedAndNamedProviders = map.get(clazz);
+        if (unNamedAndNamedProviders == null) {
+          unNamedAndNamedProviders = new UnNamedAndNamedProviders<>();
+          mapClassesToAllProviders.put(clazz, unNamedAndNamedProviders);
```
mmadev: you may need to synchronized `mapClassesToAllProviders`. the `map` here may only synchronized `mapClassesToUnBoundProviders` not `mapClassesToAllProviders`

toothpick-runtime/src/main/java/toothpick/ScopeImpl.java
```diff
@@ -139,6 +141,7 @@ public String toString() {
       builder.append(aClass.getName());
       builder.append(',');
     }
+
```
mmadev: synchronized the `mapClassesToUnBoundProviders` above

toothpick-runtime/src/main/java/toothpick/ScopeImpl.java
```diff
@@ -324,21 +325,15 @@ private void installModule(Module module) {
       map = mapClassesToUnBoundProviders;
```
mmadev: synchronized the `map`

toothpick-runtime/src/main/java/toothpick/util/ConcurrentIdentityHashMap.java
mmadev: remove it if you don't need it anymore

toothpick-runtime/src/main/java/toothpick/ScopeImpl.java
stephanenicolas: that was a bug !

toothpick-runtime/src/main/java/toothpick/ScopeImpl.java
stephanenicolas: done. thx

toothpick-runtime/src/main/java/toothpick/ScopeImpl.java
stephanenicolas: For this one I understand the concern, we get, we should synchronize, but here is the point : it never never never crashes. Do we get inconsistent data ? I don't think so, but it seems to me the bucketing should fail and crash, but I don't understand why, it passes.

PR #48, opened 2016-05-02
Installed modules scope bug

dlemures:
UPDATE : DO NOT MERGE. I THINK THIS PR IS DOING THE SAME MISTAKE AS LAST TIME...

Solved issue about classes bound in a parent scope not being injected using the right scope.

Also updated the internal provider structure and naming:
What is a internal scoped provider? A provider that is attached to one scope AND uses always the same scope create instances?

That's not true anymore, as bound classes (through a module installation) are attached to one scope BUT DO NOT always use the same scope to create instances.

So from now on, all internals providers use the class _InternalProviderImpl_ and they are considered scoped if they are attached to one scope. That happens for 2 cases:
1. Bound classes
2. Injected classes not bound but with scope instances annotations


stephanenicolas:
I see so the problem is that we could be binding providers in a scope via a module, but this is actually an unscoped provider in the old terminology... right ? It would always create instances in the scope it was bound to but it should actually use the scope it is used in...

toothpick-runtime/src/main/java/toothpick/InternalProviderImpl.java
```diff
@@ -89,7 +90,8 @@ public synchronized T get(Scope scope) {
         return instance;
       }
       if (providerFactory.hasScopeAnnotation()) {
-        providerInstance = providerFactory.createInstance(scope);
+        Scope targetScope = providerFactory.getTargetScope(scope);
+        providerInstance = providerFactory.createInstance(targetScope);
```
stephanenicolas: I don't understand these changes. The first line of the factory will do exactly that... The scope param of createInstance is immediatly assigned to the target scope of the factory. How can it solve anything ?

toothpick-runtime/src/main/java/toothpick/ScopeImpl.java
```diff
@@ -272,69 +270,60 @@ private void installModule(Module module) {
...
-    public void setMapNameToProvider(Map<String, ScopedProviderImpl<? extends T>> mapNameToProvider) {
-      this.mapNameToProvider = mapNameToProvider;
-    }
-
-    public void setUnNamedProvider(ScopedProviderImpl<? extends T> unNamedProvider) {
-      this.unNamedProvider = unNamedProvider;
+    private void putNamedProvider(String name, InternalProviderImpl<? extends T> provider) {
+      if (mapNameToProvider == null) {
+        mapNameToProvider = new HashMap<>();
+      }
```
stephanenicolas: ok this will need quite some work after the lock free PR. To make it easier, do not lazy the creation of the map.

toothpick-runtime/src/main/java/toothpick/ScopeImpl.java
```diff
@@ -161,22 +161,22 @@ private void installModule(Module module) {
...
       case INSTANCE:
-        return new ScopedProviderImpl<>(binding.getInstance());
+        return new InternalProviderImpl<>(binding.getInstance());
       case PROVIDER_INSTANCE:
         //to ensure providers do not have to deal with concurrency, we wrap them in a thread safe provider
-        return new ScopedProviderImpl<>(binding.getProviderInstance(), false);
+        return new InternalProviderImpl<>(binding.getProviderInstance(), false);
       case PROVIDER_CLASS:
-        return new ScopedProviderImpl<>(this, binding.getProviderClass(), true);
+        return new InternalProviderImpl<>(binding.getProviderClass(), true);
```
stephanenicolas: I wonder if the only problem was not here. The providers created out of bindings are indeed not scoped.

toothpick-runtime/src/main/java/toothpick/ScopeImpl.java
```diff
@@ -195,34 +195,34 @@ private void installModule(Module module) {
...
+   * The returned provider will be scoped by the public methods to use the current scope.
    */
-  private <T> UnScopedProviderImpl<? extends T> getProviderInternal(Class<T> clazz, String name) {
+  private <T> InternalProviderImpl<? extends T> getProviderInternal(Class<T> clazz, String name) {
     if (clazz == null) {
       throw new IllegalArgumentException("TP can't get an instance of a null class.");
     }
     synchronized (clazz) {
-      ScopedProviderImpl<? extends T> scopedProvider = getScopedProvider(clazz, name);
+      InternalProviderImpl<? extends T> scopedProvider = getScopedProvider(clazz, name);
```
stephanenicolas: I suggest we rename it to getBoundProvider. And in the old logic, it should have returned a UnScopedProvider, not a ScopedPovider. In the case we keep Scoped & UnScoped providers.

toothpick-runtime/src/main/java/toothpick/ScopeImpl.java
```diff
@@ -272,69 +270,60 @@ private void installModule(Module module) {
         return unNamedAndNamedProviders.unNamedProvider;
       }
 
-      Map<String, ScopedProviderImpl<? extends T>> mapNameToProvider = unNamedAndNamedProviders.getMapNameToProvider();
-      if (mapNameToProvider == null) {
+      if (unNamedAndNamedProviders.mapNameToProvider == null) {
         return null;
       }
-      return mapNameToProvider.get(bindingName);
+      return unNamedAndNamedProviders.mapNameToProvider.get(bindingName);
```
stephanenicolas: add a get method same as put.

stephanenicolas: but all those changes will make more conflicts with the lock free branch. If Unnamed provider could remain mostly unchanged, it would be easier.

toothpick-runtime/src/main/java/toothpick/ScopeImpl.java
stephanenicolas: the API is better, I agree though

stephanenicolas:
Good PR. Just a few comments... the Scope scope = factory.getTargetScope is probably useless at the beginning of the PR. Merge it when done, I will have to deal with the changes..bouhh :)

Good catch though. :)

stephanenicolas:
UPDATE : DO NOT MERGE. I THINK THIS PR IS DOING THE SAME MISTAKE AS LAST TIME...

If we have  a module that is installed in the application scope : `bind(IFoo.class).to(Foo.class)`, it is *normal* that all instances of Foo will be created in the application scope. This is not a bug, this is the expected behavior. 
Otherwise :
* Why would it be installed in the application scope otherwise ? 
* What is the difference between installing it in this scope or a lower scope ? 
* What happens if Foo needs  a dependency in a a sub-scope ?
* What is the relationship with scope annotation annotated classes ?

Let's have a chat tomorrow.

PR #52, opened 2016-05-06
Sni/cycle check multi thread

stephanenicolas:
Protect cycle detection in multi-thread. This implies a slow down in TP in debug configuration.
Make the API more dev friendly to change configuration.


dlemures:
👍

PR #63, opened 2016-05-11
clean up and + or - boosts internal provider mapping

stephanenicolas:
But still overall performance is good : 

we have on nimble droid
600 injections = 30 ms
6000 injections = 100 ms.
Our large app is probably somewhere in the middle at startup. I would say 1000 at most.
For dagger 2 it would be
600 injections = 1 ms
6000 injections = 7ms

The speed is really nice. let's say we afford 50 ms of injection time to get good testing, scope check, a funny API, MVP, flows... On average for higher numbers, the implementation is ~15 times slower than dagger. 4 times better than RG. On nexus 5, the ratio is <8 times slower than dagger 2 and 8 times faster than RG...

### DIFF
--- a/toothpick-runtime/src/main/java/toothpick/ScopeImpl.java
+++ b/toothpick-runtime/src/main/java/toothpick/ScopeImpl.java
@@ -31,8 +31,9 @@ import static java.lang.String.format;
 public class ScopeImpl extends ScopeNode {
   private static final String LINE_SEPARATOR = System.getProperty("line.separator");
 
-  private static IdentityHashMap<Class, UnNamedAndNamedProviders> mapClassesToUnBoundProviders = new IdentityHashMap<>();
-  protected IdentityHashMap<Class, UnNamedAndNamedProviders> mapClassesToAllProviders = new IdentityHashMap<>();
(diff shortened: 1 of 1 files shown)


toothpick-runtime/src/main/java/toothpick/ScopeImpl.java
```diff
@@ -31,8 +31,9 @@
 public class ScopeImpl extends ScopeNode {
   private static final String LINE_SEPARATOR = System.getProperty("line.separator");
 
-  private static IdentityHashMap<Class, UnNamedAndNamedProviders> mapClassesToUnBoundProviders = new IdentityHashMap<>();
-  protected IdentityHashMap<Class, UnNamedAndNamedProviders> mapClassesToAllProviders = new IdentityHashMap<>();
+  private static IdentityHashMap<Class, InternalProviderImpl> mapClassesToUnNamedUnBoundProviders = new IdentityHashMap<>();
```
dlemures: mapClassesToUnNamedUnBoundProviders

unbounds are always unnamed.

dlemures:
👍

PR #65, opened 2016-05-11
Introducing ReadWriteLock

dlemures:


dlemures:
It seems that the cost of locking/unlocking > that waiting for a get.

PR #75, opened 2016-05-16
Sni/configuration with class loading using reflection

stephanenicolas:
Make default configuration with reflection to load factories and member injectors.


toothpick-runtime/src/main/java/toothpick/Configuration.java
```diff
@@ -4,100 +4,182 @@
...
+      return FactoryRegistryLocator.getFactoryUsingReflection(clazz);
+    }
+
+    @Override
+    public <T> MemberInjector<T> getMemberInjector(Class<T> clazz) {
+      return MemberInjectorRegistryLocator.getMemberInjectorUsingReflection(clazz);
+    }
+  }
 
+  private static class DevelopmentConfiguration extends Configuration {
```
dlemures: This can extend BaseConfiguration. So getFactory and getMemberInjector are already defined.

toothpick-runtime/src/main/java/toothpick/Configuration.java
```diff
@@ -4,100 +4,182 @@
...
-   *
-   * @param configuration the configuration to use
-   */
-  static void setConfiguration(Configuration configuration) {
-    instance = configuration;
+  private static class ProductionConfiguration extends BaseConfiguration {
+    @Override
+    public <T> Factory<T> getFactory(Class<T> clazz) {
+      return FactoryRegistryLocator.getFactoryUsingReflection(clazz);
+    }
```
dlemures: why this method?

toothpick-runtime/src/main/java/toothpick/Configuration.java
```diff
@@ -4,100 +4,182 @@
...
+  public abstract <T> Factory<T> getFactory(Class<T> clazz);
+
+  public abstract <T> MemberInjector<T> getMemberInjector(Class<T> clazz);
+
+  /**
+   * Allows to pass custom configurations.
+   *
+   * @param configuration the configuration to use
+   */
+  static void setConfiguration(Configuration configuration) {
```
dlemures: should this be public?

toothpick-runtime/src/main/java/toothpick/Configuration.java
```diff
@@ -4,100 +4,182 @@
...
+/**
+ * Strategy pattern that allows to change various behaviors of Toothpick.
+ * The default configuration is {@link #production()}.
+ * A custom configuration can be created and used by toothpick,
+ * it is even possible to use a composition of the built-in configurations.
+ */
 public abstract class Configuration {
 
-  /*package-private.*/ static Configuration instance;
+  public static Configuration instance = production();
```
dlemures: should this be private?

dlemures:
Some comments and then 👍

toothpick-runtime/src/main/java/toothpick/Configuration.java
stephanenicolas: I don't understand your question. Production is the default configuration and the PR is about not using registries by default.

toothpick-runtime/src/main/java/toothpick/Configuration.java
stephanenicolas: +1

toothpick-runtime/src/main/java/toothpick/Configuration.java
stephanenicolas: we can't make it package private anymore as it is used internally from other packages...

toothpick-runtime/src/main/java/toothpick/Configuration.java
dlemures: BaseConfiguration already have the same method with the same implementation

toothpick-runtime/src/main/java/toothpick/Configuration.java
dlemures: Oh I see. I thought that it was in the same package.

stephanenicolas:
RTM when green

Issue #72, opened 2016-05-14
Field masking


Macarse:
Interesting!
https://github.com/stephanenicolas/toothpick/wiki/Factories-and-Member-Injectors

Is it just a performance issue or things might be broken?
Is there a way to detect and log a warning?

Macarse:
I was discussing with @dlemures in chat and we reached to the conclusion that it would not break.

```
public class A {
    String a;

    public static class B extends A {
        int a;

    }

    protected void test() {
        B b = new B();
        ((A) b).a = "hi";
        System.out.println("B.a " + b.a);
        System.out.println("B.a casted" + ((A) b).a);
    }

    public static void main(String[] args) {
      B b = new B();
      b.test();
    }
}
```

MemberInjectors will have the instance casted so there would not be a problem.

Not sure about the possibility of detection. I would guess it's impossible without adding more logic to the annotation processor.

stephanenicolas:
I don't think this work, I got quite a few things to do today, @dlemures , please add a test in the code base. Whatever the result is. But no, casting is not enough, casting doesn't transform objects.

stephanenicolas:
@Macarse please review and close. Thx !

dlemures:
Hey didn't see the comment! Thx for the PR

PR #76, opened 2016-05-16
solve issue 72

stephanenicolas:
Solves #72 

Tabarwet, ça m'en bouche un coin ! I feel really stupid but I was a 100 % sure field masking couldn't be bypassed by mere casting. I guess I was thinking more of the case where member in parent is private. But  yeah, wow ! I am gonna get asleep much brighter tonight...
:)


Macarse:
hehe. @dlemures told it would work and I treat him as a lunatic.

dlemures:
✌🏼

Issue #77, opened 2016-05-16
ToothPick --> Toothpick


stephanenicolas:
rename everywhere, in code and doc. I think we will have to checkout the wiki locally and fix all typos, the source code will be easier to refactor.

PR #83, opened 2016-05-25
Rename ToothPick to Toothpick

stephanenicolas:
Solves #77


toothpick-runtime/src/main/java/toothpick/ToothPick.java
```diff
@@ -10,14 +10,14 @@
  * <em/>A soon as you use {@code new Foo}, in a provider or a binding for instance, TP is not responsible for injecting Foo;
  * developers have to manually inject the instances they create.</em> <br/
  */
-public final class ToothPick {
+public final class Toothpick {
```
dlemures: rename file to Toothpick.java

dlemures:
👍

PR #93, opened 2016-06-20
Renaming binding methods

dlemures:
After some feedback it seems a little confusing to use the same method name for instance-class.

WDYT?


stephanenicolas:
Why not ?... RTM

PR #97, opened 2016-06-22
Renaming Smoothie Modules

dlemures:
ApplicationModule or ActivityModule are names that can be used in any app. So better to have a more specific names.


stephanenicolas:
RTM

PR #100, opened 2016-06-22
Changing Factory missing error message

dlemures:


stephanenicolas:
why don't we mention a registry setup problem any more ?

dlemures:
These problem will mainly arise when annotations are not used properly. So I'm giving more importance to that case.

@stephanenicolas I re-added the Registry case, but specifying that it is optional.

stephanenicolas:
RTM

PR #115, opened 2016-06-29
Fixing Java 7 issues with Toothpick-Compiler

dlemures:
Fixing Java 7 issues with Toothpick-Compiler


stephanenicolas:
I added a mention to toothpick, FAQ that it will works only for JDK 7+

Issue #120, opened 2016-07-12
Should we repackage the jsr 330 annotations?


stephanenicolas:
https://github.com/javax-inject/javax-inject/issues/31

stephanenicolas:
https://code.google.com/p/android/issues/detail?id=209084

johncarl81:
@stephanenicolas, I'm curious, why do you see this necessary?

stephanenicolas:
@johncarl81 the build tools actually keep every annotation class that is retained at runtime in the first dex. *And also all classes using them !* At GRPN, we do have a too large first dex and want to strip it down to essential. 

So we need to repackage the annotations classes of JSR 330 to get a more limited retention. 
See https://github.com/javax-inject/javax-inject/issues/31

@johncarl81 can you provide a link (in a new issue) to help us pas the JSR 330 TCK ?

stephanenicolas:
Released in RC6. Just as follow up, at Groupon, we use the new TP annotations for production code and the usual JSR annotations for tests, this little trick is possible easily with gradle. We use reflection for tests to perform some injections and mocking, but that is not really per say linked to TP itself.

PR #121, opened 2016-07-12
use retention class javax annotations

stephanenicolas:
#120


stephanenicolas:
After talking with @dlemures : RTM

PR #124, opened 2016-07-15
spliting registry getter method length to avoid warning

dlemures:
For big project, with a lot of factories and member injectors, we generate huge factories. That makes ART complain about the `getFactory`, `getMemberInjection` length and they don't get compiled:

```java
Method exceeds compiler instruction limit: 19072 in toothpick.Factory com.groupon.FactoryRegistry.getFactory(java.lang.Class)
Method exceeds compiler instruction limit: 19408 in toothpick.MemberInjector com.groupon.MemberInjectorRegistry.getMemberInjector(java.lang.Class)
```

Splitting the getter methods into buckets using the same approach as a HashMap:

* Use HashCode
* Create power-of-two buckets and use the last bits of the hash code to classify the factories/member injectors.

### DIFF
--- a/toothpick-compiler/src/test/java/toothpick/compiler/factory/FactoryRegistryTest.java
+++ b/toothpick-compiler/src/test/java/toothpick/compiler/factory/FactoryRegistryTest.java
@@ -6,6 +6,7 @@ import java.util.Arrays;
 import java.util.Collections;
 import javax.tools.JavaFileObject;
 import org.junit.Test;
+import toothpick.compiler.registry.generators.RegistryGeneratorTestUtilities;
 
 import static com.google.common.truth.Truth.assert_;
 import static com.google.testing.compile.JavaSourceSubjectFactory.javaSource;
@@ -24,6 +25,7 @@ public class FactoryRegistryTest {
         "package toothpick;", //
         "", //
         "import java.lang.Class;", //
+        "import java.lang.String;", //
         "import toothpick.registries.factory.AbstractFactoryRegistry;", //
         "", //
         "public final class FactoryRegistry extends AbstractFactoryRegistry {", //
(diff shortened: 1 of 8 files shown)


toothpick-compiler/src/test/java/toothpick/compiler/factory/FactoryRegistryTest.java
```diff
@@ -24,14 +24,26 @@
...
+        "import java.lang.String;", //
         "import toothpick.registries.factory.AbstractFactoryRegistry;", //
         "", //
         "public final class FactoryRegistry extends AbstractFactoryRegistry {", //
         "  public FactoryRegistry() {", //
         "  }", //
         "", //
         "  public <T> Factory<T> getFactory(Class<T> clazz) {", //
-        "    switch(clazz.getName().replace('$','.')) {", //
+        "    String className = clazz.getName().replace('$','.');", //
```
stephanenicolas: I thought we had killed the replace ... strange

toothpick-compiler/src/test/java/toothpick/compiler/factory/FactoryRegistryTest.java
```diff
@@ -24,14 +24,26 @@
...
+        "    int bucket = (className.hashCode() & 0);", //
```
stephanenicolas: = 0 ? no ?

stephanenicolas:
RTM with more testing. Handle the 0 and 1+ bucket cases

toothpick-compiler/src/test/java/toothpick/compiler/factory/FactoryRegistryTest.java
dlemures: Nope 😋

toothpick-compiler/src/test/java/toothpick/compiler/factory/FactoryRegistryTest.java
dlemures: & 0 is fine for this case:

just 1 bucket --> all classes go to the same bucket

PR #130, opened 2016-08-04
Fixes wrong binding for WifiManager.

kelsos:
It seems that in the ApplicationModule the WINDOW_SERVICE was accidentally bound to the WifiManager.


stephanenicolas:
Pretty cool @kelsos . We gonna work again on TP to polish it and release it more officially in the following days. Your fix is appreciated.

PR #132, opened 2016-08-05
Roboguice smoothie

eygraber:
@stephanenicolas I'm messing around with creating a suite of classes that provide similar functionality to the Roboguice Android component classes (`RoboActivity`, `RoboIntentService`, etc...). 

What do you think about this effort? I've also been updating smoothie class name to use `Toothpick` instead of `Smoothie` as a class prefix.


stephanenicolas:
Hi @eygraber , sorry for the long delay before checking your PR. 

Actually, it has been a real error in RG to provide the base classes, for different reasons: 
* there are mostly duplicates between each other
* there are not interesting to maintain, and take time
* users/devs can copy a sample quickly to achieve the same
* there are many base classes possible, and it's not possible to cover all them, I mean not only in the android sdk, but a lot of (bad) libs also rely on inheritance to be used and this makes a huge total of possible base classes.
* it can give the feeling to a quick visitor that TP relies on inheritance of some base classes to work, and it's not the case. 

So we won't go in this direction.

PR #137, opened 2016-08-18
Lazy & Provider issues

dlemures:
Solving 2 issues related to Lazy and Provider:

1) **IMPROVEMENT**

When Lazies dont include the type parameter, we just get an internal exception (OutOfIndex...):
```java
@Inject Lazy lazy;
@Inject Provider provider;

java.lang.IndexOutOfBoundsException: Index: 0, Size: 0
	at com.sun.tools.javac.util.List.get(List.java:476)
	at toothpick.compiler.common.ToothpickProcessor.getKindParameter(ToothpickProcessor.java:422)
	at toothpick.compiler.common.ToothpickProcessor.getInjectedType(ToothpickProcessor.java:277)
...
```

Checking for Lazy & Provider validity and displaying a proper error:
```java
Parameter foo in method/constructor test.TestMethodInjection#m is not a valid Lazy or Provider.
```

2) **BUG**

Lazies and Providers injected through constructors or methods are wrongly generated for Factories and Member Injectors. Wrong type parameter:
```java
public class A {
    @Inject
    public A(@Lazy<B> b) {
        ...
    }

    @Inject
    public void init(@Provider<C> c) {
        ...
    }
}
```
Generates:
```java
// Code generated for Lazy (Factory and Member Injector)
Lazy<T> param1 = scope.getLazy(B.class)

// Code generated for Provider (Factory and Member Injector)
Provider<T> param1 = scope.getProvider(C.class)
```

Changed to use proper type parameter:
```java
Lazy<B> param1 = scope.getLazy(B.class)
Provider<C> param1 = scope.getProvider(C.class)
```


toothpick-compiler/src/main/java/toothpick/compiler/common/ToothpickProcessor.java
```diff
@@ -234,8 +243,28 @@ protected boolean isValidInjectedType(Element injectedTypeElement) {
...
+  }
+
+  private boolean isValidProviderOrLazy(Element element) {
+    DeclaredType declaredType = (DeclaredType) element.asType();
+
+    // Contains type parameter
+    if (declaredType.getTypeArguments().isEmpty()) {
+      Element enclosingElement = element.getEnclosingElement();
+      if (enclosingElement instanceof TypeElement) {
+        error(element, "Field %s#%s is not a valid Lazy or Provider.",
```
saguinav: Instead of hardcoding "Lazy or Provider" the type should be printed as in Line 236.

mmadev:
RTM

toothpick-compiler/src/main/java/toothpick/compiler/common/ToothpickProcessor.java
dlemures: very good point.

PR #140, opened 2016-08-22
Configuration refactor

dlemures:
We have 2 different setting variables: runtime checks and reflection usage.

However, we currently have 3 configurations that are not very clear:

Configuration | Runtime checks | Reflection
------------ | ------------- | ------------
Development | X | X
Production |  | X
Reflection free |  | 

Moreover, we might want to avoid reflection (using Registries), but have runtime checks for debug builds.

So, this PR is a refactor from:
```java
Configuration.development()
Configuration.production()
Configuration.reflectionFree()
```

to:
```java
Configuration.forDevelopment()  // Default to reflection
Configuration.forProduction()  // Default to reflection
Configuration.forDevelopment().enableReflection()
Configuration.forDevelopment().disableReflection()
Configuration.forProduction().enableReflection()
Configuration.forProduction().disableReflection()
```

Where development means to include runtime checks, and production the opposite.
This way both variables are totally independent.

By default, production (no runtime checks) and reflection are used, the same as now.
Also the same style is kept.


toothpick-runtime/src/main/java/toothpick/configuration/Configuration.java
```diff
@@ -0,0 +1,105 @@
...
+ * Strategy pattern that allows to change various behaviors of Toothpick.
+ * The default configuration is {@link #production()} and {@link #reflection()}.
+ * A custom configuration can be created and used by toothpick,
+ * it is even possible to use a composition of the built-in configurations.
+ */
+public class Configuration {
+
+  public static Configuration instance = new Configuration();
+
+  private ReflectionConfiguration reflectionConfiguration = new ReflectionOnConfiguration();
```
mmadev: what is the reason we don't use reflection free by default? so TP run faster.

dlemures: It is a design decision we took. We can discuss it anyway.

If we use the reflectionfree configuration, some extra steps are needed (setting up registries, sending some annotation processing parameters).

So the default one is the one that will just run with no settings.

toothpick-runtime/src/main/java/toothpick/configuration/Configuration.java
```diff
@@ -0,0 +1,105 @@
...
+
+  /**
+   * Set reflection mode.
+   * It is slower than {@link #reflectionFree()} but it does not
+   * need any additional setup of the annotation processors.
+   * It can be used in production or development.
+   *
+   * @return a configuration set up to use reflection.
+   */
+  public Configuration reflection() {
```
saguinav: IMO something like ```enableReflection()``` would be more clear to understand.

toothpick-runtime/src/main/java/toothpick/configuration/Configuration.java
```diff
@@ -0,0 +1,105 @@
...
+
+  /**
+   * Set reflection free mode.
+   * It is faster than {@link #reflection()} but it needs
+   * some additional setup of the annotation processors.
+   * It can be used in production or development.
+   *
+   * @return an optimized reflection free configuration.
+   */
+  public Configuration reflectionFree() {
```
saguinav: IMO something like ```disableReflection()``` would be more clear to understand.

toothpick-runtime/src/main/java/toothpick/configuration/Configuration.java
```diff
@@ -0,0 +1,105 @@
...
+   * The checks performed are:
+   * <ul>
+   * <li>cycle detection: check that not 2 classes depend on each other. Note that if of them uses a Lazy instance
+   * of the other or a Producer, then there is no such cycle.</li>
+   * <li>illegal binding detection: check no scope annotated class is used as the target of a binding.</li>
+   * </ul>
+   *
+   * @return a development configuration.
+   */
+  public static Configuration development() {
```
saguinav: What about ```Configuration.forDevelopment()```?

toothpick-runtime/src/main/java/toothpick/configuration/Configuration.java
```diff
@@ -0,0 +1,105 @@
...
+  }
+
+  /**
+   * Performs no runtime checks. This configuration
+   * is faster than {@link #development()}.
+   * It can be used in production.
+   *
+   * @return a production configuration.
+   */
+  public static Configuration production() {
```
saguinav: What about ```Configuration.forProduction()```?

toothpick-runtime/src/main/java/toothpick/configuration/ReflectionOffConfiguration.java
```diff
@@ -0,0 +1,18 @@
+package toothpick.configuration;
+
+import toothpick.Factory;
+import toothpick.MemberInjector;
+import toothpick.registries.FactoryRegistryLocator;
+import toothpick.registries.MemberInjectorRegistryLocator;
+
+class ReflectionOffConfiguration implements ReflectionConfiguration {
```
saguinav: ```NoReflectionConfiguration``` ?

saguinav: ```NoReflectionConfiguration``` ?

toothpick-runtime/src/main/java/toothpick/configuration/ReflectionOnConfiguration.java
```diff
@@ -0,0 +1,18 @@
+package toothpick.configuration;
+
+import toothpick.Factory;
+import toothpick.MemberInjector;
+import toothpick.registries.FactoryRegistryLocator;
+import toothpick.registries.MemberInjectorRegistryLocator;
+
+class ReflectionOnConfiguration implements ReflectionConfiguration {
```
saguinav: ```DefaultReflectionConfiguration``` ?

toothpick-runtime/src/main/java/toothpick/configuration/Configuration.java
```diff
@@ -0,0 +1,105 @@
...
+import toothpick.MemberInjector;
+import toothpick.config.Binding;
+
+/**
+ * Strategy pattern that allows to change various behaviors of Toothpick.
+ * The default configuration is {@link #production()} and {@link #reflection()}.
+ * A custom configuration can be created and used by toothpick,
+ * it is even possible to use a composition of the built-in configurations.
+ */
+public class Configuration {
```
saguinav: final

saguinav: Implement both interfaces ```RuntimeConfiguration``` and ```ReflectionConfiguration```

toothpick-runtime/src/main/java/toothpick/configuration/Configuration.java
mmadev: I guess you already have documents somewhere. Just make sure other users aware of this option, so they can have the speed benefits.

toothpick-runtime/src/main/java/toothpick/configuration/Configuration.java
dlemures: done

toothpick-runtime/src/main/java/toothpick/configuration/Configuration.java
dlemures: done

toothpick-runtime/src/main/java/toothpick/configuration/Configuration.java
dlemures: actually it is not configuration for development. It is returning development like:
```java
Toothpick.setConfiguration(Configuration.development())
```

toothpick-runtime/src/main/java/toothpick/configuration/Configuration.java
dlemures: same

toothpick-runtime/src/main/java/toothpick/configuration/Configuration.java
dlemures: good point

dlemures: regarding the final, the user should be able to extend Configuration and provide a custom one.

toothpick-runtime/src/main/java/toothpick/configuration/ReflectionOnConfiguration.java
dlemures: Maybe we can discuss this namings later.

I'd prefer to use:
IReflectionConfiguration
NoReflectionConfiguration
ReflectionConfiguration

But we don't use I prefix for interfaces right now. We could wait for @stephanenicolas to be back and discuss about it.

I'd prefer not to use Default... As an implementation doesn't know about what is default and say nothing about what it does.

toothpick-runtime/src/main/java/toothpick/configuration/Configuration.java
dlemures: Yes it is stated.
But probably you are right and we could add more info about it as it is one of the main gains of TP.

dlemures: Issue to track it: https://github.com/stephanenicolas/toothpick/issues/141

toothpick-runtime/src/main/java/toothpick/configuration/RuntimeCheckOnConfiguration.java
```diff
@@ -0,0 +1,99 @@
...
+      if (annotation.annotationType().isAnnotationPresent(javax.inject.Scope.class)) {
+        throw new IllegalBindingException(format("Class %s cannot be bound. It has an scope annotation", clazz.getName()));
+      }
+    }
+  }
+
+  @Override
+  public void checkCyclesStart(Class clazz, String name) {
+    final Pair pair = new Pair(clazz, name);
+    final LinkedHashSet<Pair> linkedHashSet = cycleDetectionStack.get();
```
dlemures: Updated to use Pairs so name is taken into account when checking cycles

toothpick-runtime/src/main/java/toothpick/configuration/ConfigurationHolder.java
```diff
@@ -0,0 +1,19 @@
+package toothpick.configuration;
+
+/**
+ * Strategy pattern that allows to change various behaviors of Toothpick.
```
saguinav: Update the documentation here.

toothpick-runtime/src/main/java/toothpick/configuration/ConfigurationHolder.java
```diff
@@ -0,0 +1,19 @@
...
+public class ConfigurationHolder {
+
+  public static Configuration configuration = new Configuration();
+
+  /**
+   * Allows to pass custom configurations.
+   *
+   * @param configuration the configuration to use
+   */
+  public static void setConfiguration(Configuration configuration) {
```
saguinav: I think that for now it's fine to get rid off the setter too.

saguinav:
LGTM

PR #148, opened 2016-09-03
Singletons not allowed to be bound with RuntimeConfiguration

philipbjorge:
It is my understanding from the docs in [Module.java](https://github.com/stephanenicolas/toothpick/blob/6e33843714412565ff010fffbaf8954a69724be0/toothpick/src/main/java/toothpick/config/Module.java) and the test case `bindToClass_shouldCreateInjectedSingletons_whenBoundClassAnnotatedSingleton` that you can bind Singleton's to a scope.

This works with the `RuntimeCheckOffConfiguration` but not the `RuntimeCheckOnConfiguration`

This PR addresses that.

-----

I see that there are a number of failing tests related to this change.
I'm happy to address those once you confirm my interpretation of @Singleton is correct and/or that the javadocs are actually up to date (I suspect they are not).

### DIFF
--- a/toothpick-runtime/src/main/java/toothpick/configuration/Configuration.java
+++ b/toothpick-runtime/src/main/java/toothpick/configuration/Configuration.java
@@ -22,7 +22,7 @@ public class Configuration implements RuntimeCheckConfiguration, ReflectionConfi
    * <ul>
    * <li>cycle detection: check that not 2 classes depend on each other. Note that if of them uses a Lazy instance
    * of the other or a Producer, then there is no such cycle.</li>
-   * <li>illegal binding detection: check no scope annotated class is used as the target of a binding.</li>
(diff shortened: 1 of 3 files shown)


stephanenicolas:
WIP

dlemures:
Hi @philipbjorge , thx for the PR and sorry for the wait. We have been a little busy lately.

You are right, the javadoc for that file is out of date. The idea was not to allow any binding including a `scope annotated` class.

However, we are changing it as we think that it is a valid use case:
Scoped annotated targets of bindings are allowed as long as the `scoped annotation` is supported by the scope where the binding is defined.

For example, this will be possible only for the root scope (root scope - singleton):
```java
@Singleton
class Y {....}
...
bind(X.class).to(Y.class);
```

What do you think about the change?

philipbjorge:
@dlemures --
Continuing discussion on https://github.com/stephanenicolas/toothpick/issues/150

Issue #118, opened 2016-07-06
Custom scope annotations are not processed by the compiler


stephanenicolas:
Singleton is processed and we create relaxed factories in case it is annotating a class.

But it's not the case for custom scope annotations (like context singleton). I don't think we can actually do that without saying that the annotation processor should intercept every annotation, which is considered a bad practice. Maybe we can have a look at Dagger 2 and see how they deal with this as they also have custom scopes working well.

stephanenicolas:
we could provide an annotation processor parameter for this.. maybe. Not sure it works.

stephanenicolas:
This is shipped in 1.0.0

PR #156, opened 2016-10-01
Solves issue 118: support custom annotations

stephanenicolas:
Solves #118 
TODO : update the wiki. A new option has been added : `toothpick_annotations` to list the custom scope annotation classes that TP should take into account.

### DIFF
--- a/toothpick-compiler/src/main/java/toothpick/compiler/factory/FactoryProcessor.java
+++ b/toothpick-compiler/src/main/java/toothpick/compiler/factory/FactoryProcessor.java
@@ -1,16 +1,15 @@
 package toothpick.compiler.factory;
 
+import java.lang.annotation.Annotation;
 import java.util.ArrayList;
 import java.util.LinkedHashMap;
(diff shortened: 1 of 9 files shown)


toothpick-compiler/src/main/java/toothpick/compiler/factory/FactoryProcessor.java
```diff
@@ -70,10 +71,23 @@
...
+  @Override
   public boolean process(Set<? extends TypeElement> annotations, RoundEnvironment roundEnv) {
 
-    readProcessorOptions();
     findAndParseTargets(roundEnv);
+    for (TypeElement annotation : annotations) {
+      if (annotation.getAnnotationsByType(Scope.class).length != 0) {
+        createFactoriesForClassesAnnotatedWith(roundEnv, annotation);
+      }
+    }
```
dlemures: It is about "findAndParseTargets".
Could you add it inside that method?

Issue #117, opened 2016-07-05
Factory code generator should strip the generic part of dependencies


stephanenicolas:
In a class using a dependency that is generics such as:

```java
public class DaoGoodsBrowseByCategory extends GrouponBaseDao<GoodsBrowseByCategory> {

    @Inject
    public DaoGoodsBrowseByCategory(ConnectionSource connectionSource, DatabaseTableConfig tableConfig) throws SQLException {
        super(connectionSource, tableConfig);
    }
```

where `DatabaseTableConfig` is 
```java
public class DatabaseTableConfig<T>
```

TP then generates the following `Factory`
```java
public final class DaoGoodsBrowseByCategory$$Factory implements Factory<DaoGoodsBrowseByCategory> {
  @Override
  public DaoGoodsBrowseByCategory createInstance(Scope scope) {
    scope = getTargetScope(scope);
    ConnectionSource param1 = scope.getInstance(ConnectionSource.class);
    DatabaseTableConfig<T> param2 = scope.getInstance(DatabaseTableConfig.class);
    DaoGoodsBrowseByCategory daoGoodsBrowseByCategory = new DaoGoodsBrowseByCategory(param1, param2);
    return daoGoodsBrowseByCategory;
  }
```

We should just strip off the `<T>` of `    DatabaseTableConfig<T> param2 = scope.getInstance(DatabaseTableConfig.class);`

stephanenicolas:
This will be released in 1.0.0

PR #157, opened 2016-10-02
solves issue #117: Make generic usage more robust with Lazy / Provider

stephanenicolas:
Solves #117 . This is a tricky PR. 
@dlemures look at the tests, TP allows some limited usages of generics, but not all things are possible.

if we have a class 
```java
class Foo<T> {..}
```

then `Lazy<Foo>` is allowed but not `Lazy<Foo<String>>`, as there was no way to generate the code that would create the lazy and make it compile.

The problem is that `getLazy(Foo.class)` cannot generate something that would be a valid `Foo<X>`, we would need to add more type parameters to `getLazy` and we don't want that as the possibilities are infinite (e.g. `Foo<X<Y,Z>, W>`, we would need 4 more params here...).

### DIFF
--- a/toothpick-compiler/src/test/java/toothpick/compiler/factory/FactoryTest.java
+++ b/toothpick-compiler/src/test/java/toothpick/compiler/factory/FactoryTest.java
@@ -444,6 +444,93 @@ public class FactoryTest extends BaseFactoryTest {
   }
 
   @Test
+  public void testNonEmptyConstructorWithGenerics() {
+    JavaFileObject source = JavaFileObjects.forSourceString("test.TestNonEmptyConstructor", Joiner.on('\n').join(//
+        "package test;", //
+        "import java.util.List;", //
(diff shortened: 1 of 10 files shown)


toothpick-compiler/src/test/java/toothpick/compiler/factory/FactoryTest.java
```diff
@@ -444,6 +444,94 @@ public void testNonEmptyConstructor_shouldFail_whenContainsInvalidProviderParame
   }
 
   @Test
+  public void testNonEmptyConstructorWithGenerics() {
+    JavaFileObject source = JavaFileObjects.forSourceString("test.TestNonEmptyConstructor", Joiner.on('\n').join(//
+        "package test;", //
+        "import java.util.List;", //
+        "import javax.inject.Inject;", //
+        "import toothpick.Lazy;", //
```
dlemures: not needed Lazy

PR #166, opened 2016-10-12
Issue 165: Updating @Named to have RUNTIME retention so it can be det…

dlemures:
…ected by ToothPickTestModule


dlemures:
It does not make any sense anymore: https://github.com/stephanenicolas/toothpick/issues/176

Issue #167, opened 2016-10-21
Detect multiple trees in the forest of scopes


stephanenicolas:
On android, it would be useful to detect when multiple root scopes are created.

This feature is an answer to async code creating a a scope that was previously closed. Typically, a callback open the activity scope after the activity was destroyed.

The PR aims to create developer tools to detect this situation during development.

stephanenicolas:
will be shipped in 1.0.1

PR #168, opened 2016-10-21
detect multiple trees in the forest of scopes

stephanenicolas:
solves #167

### DIFF
--- a/toothpick-runtime/src/main/java/toothpick/Toothpick.java
+++ b/toothpick-runtime/src/main/java/toothpick/Toothpick.java
@@ -8,15 +8,18 @@ import toothpick.configuration.ConfigurationHolder;
  * Main class to access toothpick features.
  * It allows to create / retrieve scopes and perform injections.
  *
- * The main rule about using TP is : <b>TP will honor all injections in the instances it creates by itself.</b><br/>
- * <em/>A soon as you use {@code new Foo}, in a provider or a binding for instance, TP is not responsible for injecting Foo;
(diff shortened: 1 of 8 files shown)


toothpick-runtime/src/main/java/toothpick/Toothpick.java
```diff
@@ -41,12 +41,14 @@ public static Scope openScopes(Object... names) {
...
       previousScope = lastScope;
-      lastScope = (ScopeNode) openScope(name);
+      lastScope = (ScopeNode) openScope(name, false);
       if (previousScope != null) {
         //if there was already such a node, we add a new child
         //but there might already be such a child, in that case
         //we use it.
         lastScope = previousScope.addChild(lastScope);
+      } else {
+        ConfigurationHolder.configuration.checkMultipleRootsInTpScopeForest(lastScope);
```
dlemures: What if instead of checking `if (previousScope != null) ` all the time, you create a normal `for loop` and handle the initial element at the beginning:

```java
public static Scope openScopes(Object... names) {
     if (names == null || names.length == 0) {
       throw new IllegalArgumentException("null scopes can't be open.");
     }
 
     ScopeNode lastScope = null;
     ScopeNode previousScope = (ScopeNode) openScope(names[0], true);
     for (int i = 1; i < names.length; i++) {
        lastScope = (ScopeNode) openScope(names[i], false);
        lastScope = previousScope.addChild(lastScope);
        previousScope = lastScope;
     }
  
     return lastScope;
   }
```

Issue #170, opened 2016-11-08
ToothPickRule not reseting as expected


dlemures:
ToothPickRule does not reset the scopes after every test properly.

dlemures:
PR https://github.com/stephanenicolas/toothpick/pull/171

dlemures:
PR merged. Closing

PR #171, opened 2016-11-08
ToothPickRule should reset after test method is finished

dlemures:
Issue: https://github.com/stephanenicolas/toothpick/issues/170


stephanenicolas:
RTM

PR #177, opened 2016-12-06
Making scope reference from lazies and providers a weak reference

dlemures:
To avoid having lazies or providers that leaks the contexts after being closed.

So:
If you try to create the instance after the scope has been closed, we fail.

### DIFF
--- a/toothpick-runtime/src/test/java/toothpick/inject/lazy/InjectionOfLazyProviderTest.java
+++ b/toothpick-runtime/src/test/java/toothpick/inject/lazy/InjectionOfLazyProviderTest.java
@@ -63,4 +63,19 @@ public class InjectionOfLazyProviderTest extends ToothpickBaseTest {
     assertThat(bar2, isA(Bar.class));
     assertThat(bar2, sameInstance(bar1));
   }
+
+  @Test(expected = IllegalStateException.class)
+  public void testLazyAfterClosingScope() throws Exception {
+    //GIVEN
+    String scopeName = "";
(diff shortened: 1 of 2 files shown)


stephanenicolas:
I am not so sure about this one. It will make it harder to detect leaks, it will make the lazies (& providers) differ from injected instances, it will allow to use objects injected within a scope when the scope is dead - which should not happen... 

Can we prefer merging the boolean (invalidation of scopes when closed) first, and this one could be added to an internal release to see how it goes. I would say weak references do have a smell..

toothpick-runtime/src/test/java/toothpick/inject/lazy/InjectionOfLazyProviderTest.java
```diff
@@ -63,4 +63,19 @@ public void testNamedInjection() throws Exception {
...
+  @Test(expected = IllegalStateException.class)
+  public void testLazyAfterClosingScope() throws Exception {
+    //GIVEN
+    String scopeName = "";
+    FooWithLazy fooWithLazy = new FooWithLazy();
+
+    //WHEN
+    Toothpick.inject(fooWithLazy, Toothpick.openScope(scopeName));
+    Toothpick.closeScope(scopeName);
+    System.gc();
```
stephanenicolas: I am surprised System.gc is so reliable..

stephanenicolas:
Ok, let's just merge it, you convinced me. :)

dlemures:
Awesome 💃🏻

Issue #178, opened 2016-12-08
Make closed scopes invalid


stephanenicolas:
Make it so that a scope can't be used after it's closed.

dlemures:
PR: https://github.com/stephanenicolas/toothpick/pull/180

PR #180, opened 2016-12-09
Adding check to ensure that scopes are not used after being closed

dlemures:
Issue: https://github.com/stephanenicolas/toothpick/issues/178

Adding check to ensure that scopes are not used after being closed.

### DIFF
--- a/toothpick-runtime/src/main/java/toothpick/ScopeImpl.java
+++ b/toothpick-runtime/src/main/java/toothpick/ScopeImpl.java
@@ -50,6 +50,7 @@ public class ScopeImpl extends ScopeNode {
 
   @Override
   public <T> T getInstance(Class<T> clazz, String name) {
+    crashIfClosed();
     ConfigurationHolder.configuration.checkCyclesStart(clazz, name);
     T t;
     try {
@@ -67,8 +68,8 @@ public class ScopeImpl extends ScopeNode {
 
   @Override
   public <T> Provider<T> getProvider(Class<T> clazz, String name) {
-    InternalProviderImpl<? extends T> provider = lookupProvider(clazz, name);
-    return new ThreadSafeProviderImpl<>(this, provider, false);
+    crashIfClosed();
+    return new ThreadSafeProviderImpl<>(this, clazz, name, false);
   }
 
   @Override
@@ -78,8 +79,8 @@ public class ScopeImpl extends ScopeNode {
 
   @Override
   public <T> Lazy<T> getLazy(Class<T> clazz, String name) {
-    InternalProviderImpl<? extends T> provider = lookupProvider(clazz, name);
-    return new ThreadSafeProviderImpl<>(this, provider, true);
+    crashIfClosed();
+    return new ThreadSafeProviderImpl<>(this, clazz, name, true);
   }
 
   @Override
@@ -500,6 +501,13 @@ public class ScopeImpl extends ScopeNode {
       } else {
         return previous;
       }
+    }
+  }
+
+  private void crashIfClosed() {
+    if (!isOpen) {
+      throw new IllegalStateException(String.format("The scope with name %s has been already closed."
+          + " It is not possible to use it to create new instances.", name));
     }
   }
 
--- a/toothpick-runtime/src/main/java/toothpick/ScopeNode.java
+++ b/toothpick-runtime/src/main/java/toothpick/ScopeNode.java
@@ -78,6 +78,7 @@ public abstract class ScopeNode implements Scope {
   //as setting the parent is called only once when creating a node
   protected final List<ScopeNode> parentScopes = new CopyOnWriteArrayList<>();
   protected Object name;
+  protected boolean isOpen = true;
   //same here for lock free access
   protected final Set<Class<? extends Annotation>> scopeAnnotationClasses = new CopyOnWriteArraySet<>();
 
@@ -259,6 +260,10 @@ public abstract class ScopeNode implements Scope {
     childrenScopes.remove(child.getName());
     //make the ex-child a new root.
     child.parentScopes.clear();
+  }
+
+  void close() {
+    isOpen = false;
   }
 
   private void checkIsAnnotationScope(Class<? extends Annotation> scopeAnnotationClass) {
--- a/toothpick-runtime/src/main/java/toothpick/Toothpick.java
+++ b/toothpick-runtime/src/main/java/toothpick/Toothpick.java
@@ -1,5 +1,6 @@
 package toothpick;
 
+import java.util.ArrayList;
 import java.util.concurrent.ConcurrentHashMap;
 import toothpick.configuration.Configuration;
 import toothpick.configuration.ConfigurationHolder;
@@ -121,7 +122,9 @@ public final class Toothpick {
    */
 
   public static void reset() {
-    MAP_KEY_TO_SCOPE.clear();
+    for (Object name : new ArrayList<>(MAP_KEY_TO_SCOPE.keySet())) {
+      closeScope(name);
+    }
     ConfigurationHolder.configuration.onScopeForestReset();
     ScopeImpl.reset();
   }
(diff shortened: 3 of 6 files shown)


dlemures:
Unit Tests pendings

toothpick-runtime/src/main/java/toothpick/Toothpick.java
```diff
@@ -121,7 +122,9 @@ public static void closeScope(Object name) {
    */
 
   public static void reset() {
-    MAP_KEY_TO_SCOPE.clear();
+    for (Object name : new ArrayList<>(MAP_KEY_TO_SCOPE.keySet())) {
+      closeScope(name);
+    }
```
dlemures: so scopes closed this way:

1. are properly closed.
2. and cannot be opened after closed either.

toothpick-runtime/src/main/java/toothpick/ScopeImpl.java
```diff
@@ -78,6 +80,7 @@ public ScopeImpl(Object name) {
 
   @Override
   public <T> Lazy<T> getLazy(Class<T> clazz, String name) {
+    crashIfInvalid();
```
stephanenicolas: as you want. Though having in the scope implementation makes more sense conceptually.

toothpick-runtime/src/main/java/toothpick/ScopeNode.java
```diff
@@ -78,6 +78,7 @@
   //as setting the parent is called only once when creating a node
   protected final List<ScopeNode> parentScopes = new CopyOnWriteArrayList<>();
   protected Object name;
+  protected boolean isValid = true;
```
stephanenicolas: isClosed ? Or isOpen ?

toothpick-runtime/src/main/java/toothpick/Toothpick.java
stephanenicolas: we should still clear the map, this is used not to leak during tests. Even though we use a weak reference, we should clear the map. The leak of scopes without references to contexts (on Android) is not much better than leaking a context. 

TP should not reference the scopes, the errors in leaks will come from devs still referencing and using them in some way.

toothpick-runtime/src/main/java/toothpick/ScopeNode.java
```diff
@@ -261,6 +262,10 @@ void removeChild(ScopeNode child) {
     child.parentScopes.clear();
   }
 
+  void invalidate() {
```
stephanenicolas: close()

stephanenicolas:
Why did you say that

>There is one exception that won't fail:
When we use a closed scope to create an instance from a lazy (or provider) that does not contain any dependency (we are just creating the instance, but it does not need to inject anything).

toothpick-runtime/src/main/java/toothpick/ScopeImpl.java
stephanenicolas: We need to, it would get rid of the problem you mention when you say:

> There is one exception that won't fail:
When we use a closed scope to create an instance from a lazy (or provider) that does not contain any dependency (we are just creating the instance, but it does not need to inject anything).

toothpick-runtime/src/main/java/toothpick/ScopeImpl.java
```diff
@@ -503,6 +506,13 @@ private void installModule(Module module) {
     }
   }
 
+  private void crashIfInvalid() {
+    if (!isValid) {
```
dlemures: yes, good point, changed to `isOpen`

toothpick-runtime/src/main/java/toothpick/ScopeNode.java
dlemures: changed to `isOpen`

toothpick-runtime/src/main/java/toothpick/ScopeNode.java
dlemures: updated thx

dlemures:
Unit tests added and comments fixed/replied

toothpick-runtime/src/main/java/toothpick/Toothpick.java
stephanenicolas: cool

stephanenicolas: But shouldn't we synchronize on something ?

toothpick-runtime/src/main/java/toothpick/ScopeImpl.java
stephanenicolas: I don't understand why you mention GC. It looks to me that the chain you describe in: 

> ThreadSafeProviderImpl.get -> InternalProviderImpl.get -> find factory using registry -> factory.createInstance -> return new ClassWithNoDependencies()

is ***not*** using any scope, hence it could be techincally possible to use a lazy after the scope is closed. Why does it matter that the scope is GC'ed or not ?

Also, the chain above looks a bit strange, a scope could not use a factory but a singleton for instance..

But I understand that the scope is not used after the lazy is created. Honestly I think we should not have any different behavior between something provided directly (as an instance) or via a provider or a lazy. I understand that for us it's different, but form a user and from a DI perspective I don't think of any reasonable meaning of having different behaviors. It looks more like an akward implementation detail (we could actually just have called scope.getInstance in the lazy). I can't make any sense out of such a deferred behavior (in time).

toothpick-runtime/src/main/java/toothpick/Toothpick.java
dlemures: It works in a similar way to `clear()`: reflecting the state of the hash table at some point.
So we clear the table with the elements available at the moment of starting the operation.

The reason to create a copy is because the iterator is meant to be used only from one thread.

toothpick-runtime/src/main/java/toothpick/ScopeImpl.java
dlemures: Updated, check last comment

dlemures:
@stephanenicolas CHANGING CORE TP LOGIC.
EXTREME REVIEW NEEDED

toothpick-runtime/src/test/java/toothpick/ScopeImplTest.java
```diff
@@ -112,6 +112,42 @@ public void testToBinding_shouldFail_whenAddingANullBinding() {
...
+    //GIVEN
+    ScopeImpl scope = new ScopeImpl("");
+
+    //WHEN
+    scope.close();
+    scope.getProvider(Foo.class);
+
+    //THEN
+  }
+
```
stephanenicolas: Damn cool.

stephanenicolas:
Honestly I would RTM based on extensive set of tests. But if you want a careful review, it will not be possible before Monday night / Tuesday for me.

toothpick-runtime/src/test/java/toothpick/ToothpickTest.java
```diff
@@ -63,7 +74,7 @@ public void openScopes_shouldFail_whenScopeNamesAreNull() throws Exception {
     Toothpick.openScopes((Object[]) null);
 
     //THEN
-    fail("Shoudl ahve thrown an exception");
+    fail("Should have thrown an exception");
```
stephanenicolas: Actually this is useless as `@Test` as an `expected` param.

toothpick-runtime/src/test/java/toothpick/ToothpickTest.java
```diff
@@ -97,6 +108,18 @@ public void destroyScope_shouldClearThisScope_WhenThisScopesWasCreated() throws
     assertThat(scopeAfterReset, not(sameInstance(scope)));
   }
 
+  @Test
+  public void destroyScope_shouldMarkThisScopeAsClosed() throws Exception {
```
stephanenicolas: closeScope...

stephanenicolas:
A few comments, but I looked at the changes, it's really fine for me. I like much better our implementation like this indeed. I find it more consistent like this and the new tests tend to demonstrate it too. 

Thx. 🚢 it

toothpick-runtime/src/test/java/toothpick/ToothpickTest.java
dlemures: removed

toothpick-runtime/src/test/java/toothpick/ToothpickTest.java
dlemures: updated

dlemures:
Merging, trying in MGA with snapshot Monday.

Issue #181, opened 2017-01-02
Code generator mistake results on compile time error while injecting static nested classes


lrn2prgrm:
I want to avoid the use of Toothpick class everywhere i want an injection (if there is another way to accomplish this please let me know) so the idea was to call the injection in the constructor so it will be executed on child's super() call.

_Current behavior errors at compile time with message on Child$$MemberInjector 
cannot find symbol. symbol: Parent$$MemberInjector_
```java
    public class Child extends Parent{
      @Inject
      SomeDependency someDependency;      
      Child () {  
        super(); 
        /* I pretend injection of someDependency to happen inside super
        * without having to call and import the Toothpick class in the child class */
      }
    }
    public class Parent{
      @Inject
      OtherDependency otherDependency;      
      public RESTController() {
        Scope scope = Toothpick.openScope("scope");
        Toothpick.inject(this, scope);
      }
    }
```
I would like to suggest this as a feature request if it was possible to implement.

stephanenicolas:
The pattern you mention is one of the possible way to use TP. I am not sure
to understand exactly what feature you request, it should just work.

Also, it's surprising that you get the error

*Child$$MemberInjectorcannot find symbol. symbol: Parent$$MemberInjector*

If Parent is accessible when compiling Child, then the Parent
MemberInjector is also accessible to the compiler when it compiles the
Child MemberInjector. I can't think of any case where this error can happen
? Can you set up a small project on GH to demonstrate the issue ?



2017-01-02 1:37 GMT+01:00 J Armando Cordova <notifications@github.com>:

lrn2prgrm:
I haven't been able to reproduce the behavior but this code never gets injected. (both dependencies print null)

`import toothpick.Scope;
import toothpick.Toothpick;
import toothpick.config.Module;

import javax.inject.Inject;

/**
 * Created by jarma on 1/2/2017.
 */
public class TPIssueMin {
	static String scope = "scope";

	public static class Dependency {@Inject Dependency() {}}
	public static class Parent {
		@Inject	Dependency parentDependency;
		Parent() { Toothpick.inject(this, Toothpick.openScope(scope)); }
	}
	public static class Child extends Parent {
		@Inject Dependency childDependency;
		Child() { super(); }
	}

	public static void main(String[] args) {
		Scope s = Toothpick.openScope(scope);
		s.installModules(new Module() {{
			bind(Dependency.class).toInstance(new Dependency());
		}});
		Child child = new Child();
[...]

PR #183, opened 2017-01-03
Issue-181: Code generated for super member injector when it is static is wrong

dlemures:
Issue: https://github.com/stephanenicolas/toothpick/issues/181


stephanenicolas:
RTM thx !

Issue #182, opened 2017-01-03
Replace is suboptimal in the registries


stephanenicolas:
We do a replace operation in the registries to lookup for the factories and member injectors, before bucketing. But this is so suboptimal: the bucket entries themselves should be replaced and the match should be direct.

![image](https://cloud.githubusercontent.com/assets/1832925/21615677/daf06842-d1de-11e6-87e4-5a4961d0460f.png)

dlemures:
PR: https://github.com/stephanenicolas/toothpick/pull/184

dlemures:
PR merged

PR #184, opened 2017-01-03
Issue-182: Removing string replace from registries and changing hardc…

dlemures:
…oded string inside switch block instead

Issue: https://github.com/stephanenicolas/toothpick/issues/182


stephanenicolas:
image

dlemures:
Adding extra test to check registries working at runtime for nested classes.

Issue #176, opened 2016-11-23
Document how to get Android instrumentation tests to run with ToothPick on API 19 or older


efung:
When trying to run an Android instrumentation test while using ToothPick, I encountered an issue on API 19 devices/emulators that took some time to resolve. Here is an [example repository](https://github.com/efung/ToothPickTest) that demonstrates the problem.

If you edit the `app/build.gradle` file to remove the exclusion of the `javax.inject` dependency from Espresso core, the tests will fail to run with the following [crash log](https://github.com/stephanenicolas/toothpick/files/609505/crash.txt). The relevant exception messages are  `java.lang.NoClassDefFoundError: toothpick/smoothie/provider/AccountManagerProvider`, caused by `java.lang.IllegalAccessError: Class ref in pre-verified class resolved to unexpected implementation`.

A previous warning message notes that `Class resolved by unexpected DEX` and that `(Ltoothpick/smoothie/provider/AccountManagerProvider; had used a different Ljavax/inject/Provider; during pre-verification)`

As far as I can understand, Dalvik (used in API 16 and 19 devices/emulators) cannot handle multiple DEX files, whereas ART (in API 21 and higher) can. So during compilation, I think both copies of `javax.inject` get included? Since both ToothPick and Espresso define/depend on `javax.inject`, probably one copy gets included and the mismatch is detected at runtime.

A similar issue exists when using Dagger and Espresso as described in these two posts:
- https://stackoverflow.com/questions/27641965/dagger-code-giving-noclassdeffounderror-when-espresso-tests-are-run-normal-run/27717382#27717382
- http://www.donnfelker.com/android-studio-espresso-2-0-classnotfoundexception/

I see from e44040f945ce1871261f425c1a92c9050c1a4e40 that ToothPick explicitly needs to define their own annotations to change their retention policy.

I'm suggesting that:
- The `toothpick-javax-annotations` module be called out more explicitly
- Provide an example application, or update the `smoothie-sample` project, or add some documentation to the Wiki, about how to use Espresso with ToothPick.

dlemures:
@efung sorry for the delay.

It is an open discussion we are having...

I think that the suggestion is a good one: if the user doesn't include `toothpick-javax-annotations` as a dependency, we use the default javax annotations.

dlemures:
PR: https://github.com/stephanenicolas/toothpick/pull/193

dlemures:
@efung we have made `toothpick-javax-annotations` optional.

[...]

PR #193, opened 2017-01-20
Issue-176: Use standard javax annotation by default

dlemures:
Issue: https://github.com/stephanenicolas/toothpick/issues/176

Tested in MGA: 
* Using default javax for testing
* Using TP javax for production, replacing with TP javax annotations

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -114,6 +114,7 @@ subprojects { project ->
 }
 
 ext.deps = [// Common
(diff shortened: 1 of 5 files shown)


dlemures:
Wiki update: https://github.com/stephanenicolas/toothpick/wiki/Toothpick-javax-Annotations

stephanenicolas:
RTM

PR #216, opened 2017-05-28
Auto close activities scopes and introduce close listeners

stephanenicolas:
Small experiment...

What is not so good is that we can easily create activity leaks with the closedListener:

```java
//method ref 
//= lambda 
//= anonymous inner class 
//= leak of this as it remains in the presenter scope.
 openScope(PRESENTER_SCOPE).setClosedListener(rxPresenter::stop); 
```

It does not actually :D 
```java
// $FF: synthetic class
final class RxMVPActivity$$Lambda$4 implements ScopeClosedListener {
    private final RxPresenter arg$1;

    private RxMVPActivity$$Lambda$4(RxPresenter var1) {
        this.arg$1 = var1;
    }

    private static ScopeClosedListener get$Lambda(RxPresenter var0) {
        return new RxMVPActivity$$Lambda$4(var0);
    }

    @Hidden
    public void onClosed() {
        this.arg$1.stop();
    }

    public static ScopeClosedListener lambdaFactory$(RxPresenter var0) {
        return new RxMVPActivity$$Lambda$4(var0);
    }
}
```

because of [retro lambda optimizing stateless lambdas to singletons](https://github.com/orfjackal/retrolambda#backported-language-features).

I checked and there is no leak. Moreover we don't need to call stop on the subject... it's just garbage collected. 

We should add this kind of stuff, maybe with a better API though ;)

### DIFF
--- a/smoothie-sample/src/main/java/com/example/smoothie/RxMVPActivity.java
+++ b/smoothie-sample/src/main/java/com/example/smoothie/RxMVPActivity.java
@@ -43,7 +43,9 @@ public class RxMVPActivity extends Activity {
     setContentView(R.layout.simple_activity);
     ButterKnife.bind(this);
     title.setText("MVP");
-    subscription = rxPresenter.subscribe(aLong -> subTitle.setText(String.valueOf(aLong)));
+    subscription = rxPresenter
+        .getTimeObservable()
+        .subscribe(aLong -> subTitle.setText(String.valueOf(aLong)));
     button.setVisibility(View.GONE);
   }
 
--- a/smoothie-sample/src/main/java/com/example/smoothie/RxMVPActivity.java
+++ b/smoothie-sample/src/main/java/com/example/smoothie/RxMVPActivity.java
@@ -18,6 +18,9 @@ import toothpick.Scope;
 import toothpick.Toothpick;
 import toothpick.smoothie.module.SmoothieActivityModule;
 
+import static toothpick.Toothpick.openScopes;
+import static toothpick.smoothie.module.Smoothie.autoCloseActivityScopeAndParentScope;
+
 public class RxMVPActivity extends Activity {
 
   public static final Class PRESENTER_SCOPE = Presenter.class;
@@ -29,10 +32,11 @@ public class RxMVPActivity extends Activity {
   @BindView(R.id.hello) Button button;
   private Subscription subscription;
 
-  @Override
-  protected void onCreate(Bundle savedInstanceState) {
-    scope = Toothpick.openScopes(getApplication(), PRESENTER_SCOPE, this);
+  @Override protected void onCreate(Bundle savedInstanceState) {
+    scope = openScopes(getApplication(), PRESENTER_SCOPE, this);
     scope.installModules(new SmoothieActivityModule(this));
+    autoCloseActivityScopeAndParentScope(this);
+
     super.onCreate(savedInstanceState);
     Toothpick.inject(this, scope);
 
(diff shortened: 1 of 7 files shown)


smoothie/src/main/java/toothpick/smoothie/module/Smoothie.java
```diff
@@ -0,0 +1,97 @@
...
+    public AutoCloseActivityParentLifecycleCallbacks(Activity activity) {
+      targetActivity = activity;
+    }
+
+    @Override public void onActivityDestroyed(Activity activity) {
+      if (activity == targetActivity) {
+        closeScope(activity);
+        if(activity.isFinishing()) {
+          final Scope parentScope = openScope(activity).getParentScope();
+          closeScope(parentScope.getName());
```
dlemures: this really MVP dependent

smoothie/src/main/java/toothpick/smoothie/module/Smoothie.java
```diff
@@ -0,0 +1,97 @@
...
+
+    public AutoCloseActivityParentLifecycleCallbacks(Activity activity) {
+      targetActivity = activity;
+    }
+
+    @Override public void onActivityDestroyed(Activity activity) {
+      if (activity == targetActivity) {
+        closeScope(activity);
+        if(activity.isFinishing()) {
+          final Scope parentScope = openScope(activity).getParentScope();
```
dlemures: the `Activity` scope is already closed, you are opening a new one with no parent

smoothie/src/main/java/toothpick/smoothie/module/Smoothie.java
```diff
@@ -0,0 +1,97 @@
...
+        }
+      }
+      final Application application = activity.getApplication();
+      application.unregisterActivityLifecycleCallbacks(this);
+    }
+  }
+
+  private static class AutoCloseActivitiesLifecycleCallbacks extends DefaultLifecycleCallbacks {
+    @Override public void onActivityDestroyed(Activity activity) {
+      closeScope(activity);
```
dlemures: this is meant to close all possible activity scopes inside ondestroy of each activity?

Probably we should add javadocs

toothpick-runtime/src/main/java/toothpick/ScopeNode.java
```diff
@@ -264,6 +265,9 @@ void removeChild(ScopeNode child) {
 
   void close() {
     isOpen = false;
+    if(this.closeListener != null) {
```
dlemures: space after if. do you need to use this here (but not in the next line)?

smoothie-sample/src/main/java/com/example/smoothie/RxMVPActivity.java
```diff
@@ -45,19 +49,11 @@ protected void onCreate(Bundle savedInstanceState) {
...
-      Toothpick.closeScope(PRESENTER_SCOPE);
-    }
     super.onDestroy();
   }
 
-  @javax.inject.Scope
-  @Target(ElementType.TYPE)
-  @Retention(RetentionPolicy.RUNTIME)
-  public @interface Presenter {}
+  @javax.inject.Scope @Target(ElementType.TYPE) @Retention(RetentionPolicy.RUNTIME)
```
dlemures: why the format change?

smoothie/src/main/java/toothpick/smoothie/module/Smoothie.java
```diff
@@ -0,0 +1,97 @@
+package toothpick.smoothie.module;
```
dlemures: i'd add javadocs, difficult to understand

toothpick-runtime/src/main/java/toothpick/ScopeNode.java
```diff
@@ -276,4 +280,9 @@ private void checkIsAnnotationScope(Class<? extends Annotation> scopeAnnotationC
   private boolean isScopeAnnotationClass(Class<? extends Annotation> scopeAnnotationClass) {
     return scopeAnnotationClass.isAnnotationPresent(javax.inject.Scope.class);
   }
+
+  @Override
+  public void setClosedListener(ScopeClosedListener closeListener) {
+    this.closeListener = closeListener;
```
dlemures: does it not lead to bad programming practices? -> Using this to know when activity is destroyed for instance.

Issue #196, opened 2017-01-31
Change ContextSingleton retention policy to RUNTIME


zawadz88:
Hi,
First of all thanks for the great library!
Recently, I tried updating toothpick from 1.0.3 to 1.0.5 in my Android project and got:
`Error:Scope Annotation toothpick.smoothie.annotations.ContextSingleton does not have RUNTIME retention policy.` during compilation.
I checked and `ContextSingleton` from _smoothie_ has runtime policy `CLASS` currently unlike e.g. Singleton from javax.inject or the custom scope in _smoothie-sample_ which probably causes this error. This is probably due to the changes in #191 

Could you help please?

dlemures:
Hi @zawadz88 ,

Thanks for the issue.

This class was meant to be there just as an example. However, the javadoc is not really clear and probably the class is not useful either.

Therefore, we will deprecate the class, change the retention and update the javadoc.
The class will be removed in a later release.

If you are using context singletons, please create the annotation yourself and bind it to the right scopes:
https://github.com/stephanenicolas/toothpick/wiki/Scope-Annotations#binding-a-scope-to-a-scope-annotation

stephanenicolas:
I would add that `ContextSingleton` was really a RoboGuice thing. It looked really cool at that time as you could quickly inject a context. At Groupon, we used a TP ContextSingleton for our migration from RG to TP. 

But, later, we just removed it. The issue is that when you `@Inject Context`, then you just don't know exactly which context you are injecting : an activity or the app context ?? And one can create a leak, not the other one, simply depending on where you inject your entity. 

So, definitely ContextSingleton is more a simple example of a custom annotation but it should not be used in a production app. At Groupon, we ActivityScope, ServiceScope, and we never inject a context, always either the Application or an Activity or a Service, so that we have full control over what is injected. And we use the scope verification mechanism of TP to be sure we don't leak the activity.

zawadz88:
Thanks for the clarification and sorry for the misclick ;)

HugoGresse:
I was searching how to inject the `Application` in a `@Singleton` is there any example on the documentation/demo? 
I also don't find any documentation about Toothpick. I'm just a beginner with DI :/

stephanenicolas:
If you install the smoothie application module in your app scope, then you can simply use `@Inject Application app` in this scope or any sub scope. 

All the docs of TP is in the wiki in the repo... @HugoGresse

HugoGresse:
[...]

PR #231, opened 2017-08-22
Address issue #196

stephanenicolas:
Addresses #196


smoothie/src/main/java/toothpick/smoothie/annotations/ContextSingleton.java
```diff
@@ -15,6 +16,6 @@
  */
 @Scope
 @Documented
-@Retention(CLASS)
```
dlemures: can't we remove this class?

dlemures: or deprecate and remove in TP 2

stephanenicolas: I will deprecate it for now.

dlemures:
RTM

Issue #232, opened 2017-08-22
Allow to reset a single scope for integration tests


stephanenicolas:
blah blah.. context. Link this PR to the original issue, explain the integration test context where the application is recycled during multiple integration tests.

TODO:
* [x] `reset` --> `resetUnBoundProviders` (this api change is not part of the public api, so version change can still be y+1 in x.y.z.
* [x] create a `reset` method that resets the internal scope state : `mapClassesToNamedBoundProviders` + `mapClassesToUnNamedBoundProviders` + `hasTestModules = false` + `isOpen = true` + `scopeAnnotationClasses`.
* [x] provide a snapshot of 1.1.0 for Cody to try.
* [x] add a wiki page to explain how to use this feature for integration tests (moved to https://github.com/stephanenicolas/toothpick/issues/255)
* [x] a sample integration test (#242)
* [x] link the wiki page from the issue where the ROOT_SCOPE workaround was first suggested. 
Subtask assigned to cody. We will add him as a TP contributor when PR is RTM.
* [x] internal to GRPN: kick off integration tests api calls mocking project + notify matthijs.

dlemures:
https://github.com/stephanenicolas/toothpick/pull/234 (1, 2 & 3)

stephanenicolas:
As an update on this topic, the test orchestrator allows to get a new app instance for every instrumentation test: https://developer.android.com/training/testing/junit-runner.html#using-android-test-orchestrator

PR #242, opened 2017-09-05
Add android test with test modules; update gradle to beta7

chenthorne:
Fix: https://github.com/stephanenicolas/toothpick/issues/232


smoothie-sample/src/androidTest/java/com/example/smoothie/SimpleActivityTest.java
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    @Test
+    public void testWithContextNamerTestModuleWithMock() {
+        ContextNamer contextNamer = EasyMock.createMock(ContextNamer.class);
+
+        expect(contextNamer.getApplicationName()).andReturn("TestAppNameMock");
+        expect(contextNamer.getActivityName()).andReturn("TestActivityNameMock");
+
+        EasyMock.replay(contextNamer);
```
stephanenicolas: you can use static imports everywhere

smoothie-sample/src/main/java/com/example/smoothie/SimpleApp.java
```diff
@@ -22,6 +22,10 @@ public void onCreate() {
     MemberInjectorRegistryLocator.setRootRegistry(new com.example.smoothie.MemberInjectorRegistry());
     FactoryRegistryLocator.setRootRegistry(new com.example.smoothie.FactoryRegistry());
 
+    initToothpick();
+  }
+
+  public void initToothpick() {
     Scope appScope = Toothpick.openScope(this);
     appScope.installModules(new SmoothieApplicationModule(this));
```
stephanenicolas: I am wondering if it would not be more clear to only install modules here and pass the scope as a param. 
What needs to be called from the test is only this, not the creation of the scope itself.

But that's really just a detail.

smoothie-sample/build.gradle
```diff
@@ -4,7 +4,7 @@ buildscript {
     google()
   }
   dependencies {
-    classpath ("com.android.tools.build:gradle:3.0.0-beta2")
+    classpath ('com.android.tools.build:gradle:3.0.0-beta4')
```
stephanenicolas: you can even go beta 6 now ;)

chenthorne:
@stephanenicolas Addressed your comments, feel free to review again

Issue #258, opened 2017-10-13
1.1.0 Generated annotation not available on Android


RocketRider:
In the Version 1.1.0  the Class "javax.annotation.Generated" can not be resolved on Android.
Relates to: #256

stephanenicolas:
Thx @RocketRider , you need to add 
```java
provided 'javax.annotation:jsr250-api:1.0'
```
(or compileOnly if you use GAP3). We will add the dependency to the generated core lib.

stephanenicolas:
@RocketRider , do you wanna give a try to our latest snapshot. It contains the PR above. Versions 1.1.1-SNAPSHOT

stephanenicolas:
We are releasing 1.1.1 which fixes the issue.

Issue #253, opened 2017-10-10
Generated classes should use the Generated annotation


stephanenicolas:
See https://github.com/frankiesardo/icepick/pull/109
This also applies to MemberInjectorRegistry

PR #260, opened 2017-10-13
Fix issue #258, #256, #253: remove dependency to JSR 250 annotations, fix missing overrides.

stephanenicolas:
Fix issue #258, #256, #253: remove dependency to JSR 250 annotations, fix missing overrides.


dlemures:
RTM

Issue #146, opened 2016-08-29
Reflection Free Configuration and Proguard


philipbjorge:
Is there any guidance on how to use the reflection free configuration with proguard and obfuscation?
Based on looking at the factories, it looks like I'll need to keep my class names at a minimum (possibly method names too).

`-keepnames class com.myapp.**`

Do you have any advice on how I can make this more specific?

philipbjorge:
Here's what I've got working on my project...

```
-keep class javax.inject.**

# Do not obfuscate annotation scoped classes
-keepnames @javax.inject.Singleton class *
# Add any custom defined @Scope (e.g. @Singleton) annotations here
# because proguard does not allow annotation inheritance rules

# Do not obfuscate classes with Injected Constructors
-keepclasseswithmembernames class * {
    @javax.inject.Inject <init>(...);
}

# Do not obfuscate classes with Injected Fields
-keepclasseswithmembernames class * {
    @javax.inject.Inject <fields>;
}

# Do not obfuscate classes with Injected Methods
-keepclasseswithmembernames class * {
    @javax.inject.Inject <methods>;
}
```

dlemures:
Hi @philipbjorge ,

Thanks for the rules!

Do you need `-keep class javax.inject.**`? 
As we used them just at compile time as far as i remember.

stephanenicolas:
As soon as @philipbjorge answers our question, we should add this to the wiki page about configurations.

philipbjorge:
@dlemures When I remove the `-keep class javax.inject.**`, I get the following error at runtime:

```
Unable to start activity ComponentInfo{com.xyz/com.xyz.activities.MainActivity}: e.c.e: No factory could be found for class com.xyz.a.a. Check that the class has either a @Inject annotated constructor or contains @Inject annotated members. If using Registries, check that they are properly setup with annotation processor arguments.
android.app.ActivityThread.performLaunchActivity
```

Any ideas why or how to debug?

philipbjorge:
Looking at my stacktraces + mappings.txt makes me think something real funny is going on.
I'll keep digging and might be able to remove the javax.inject rule...

philipbjorge:
OK, so I had the following rule to keep the names of all the classes annotated with my custom scope annotations
`-keepnames @com.xyz.support.di.scopes.* class *`

Turns out that doesn't work and you need to list them all out manually
```
-keepnames @javax.inject.Singleton class *
[...]

PR #273, opened 2018-01-07
Generate registries in a way that works with obfuscation

pshmakov:
With these changes it's possible to use obfuscation with Toothpick provided the following:
1) -adaptclassstrings added to proguard rules
2) support_obfuscation : "true" added to annotation processing arguments.
3) Configuration with registries is used.

The issue is described here:
https://github.com/stephanenicolas/toothpick/issues/146

Switching on class name strings doesn't work after -adaptclassstrings is applied because the hashcodes of strings are calculated and written into bytecode before the obfuscation.
I replaced switch with a HashMap. The changes take effect only if support_obfuscation : "true" was set.


PiotrBandurski:
@stephanenicolas any info when this pr will be merged?

dimsuz:
Me and few people I know also hoping for this to get reviewed and merged. Obfuscation is rather important in some projects.

dimsuz:
@pshmakov 

```
-adaptclassstrings added to proguard rules
```
I guess this PR should add this to `README.md` then?

pshmakov:
@dimsuz 
Readme.md contains only general information. It's the wiki that would need updating, but it's not in the repo itself.

Gaket:
@stephanenicolas, we wanted to bring Toothpick into a banking app but the lack of Proguard is a show-stopper for us. When will you have some time to look at this PR? We would like to know if we should wait for your decision or start the project with Dagger.

stephanenicolas:
Sorry about the PR not being merged. I am adding an entry in my calendar to
make sure it's done this week.

Thanks for the follow up and sorry for the delay, we have been quite busy
recently with Dart & Henson 3...

S.


2018-03-04 3:18 GMT-08:00 Artur Badretdinov <notifications@github.com>:

toothpick-compiler/src/test/java/toothpick/compiler/memberinjector/ObfuscationFriendlyMemberInjectorRegistryTest.java
```diff
@@ -0,0 +1,151 @@
...
+                    "import java.lang.String;",
+                    "import java.util.HashMap;",
+                    "import java.util.Map;",
+                    "import toothpick.registries.memberinjector.AbstractMemberInjectorRegistry;",
+                    "",
+                    "public final class MemberInjectorRegistry extends AbstractMemberInjectorRegistry {",
+                    "    private final Map<String, MemberInjector> factories = new HashMap<>();",
+                    "",
+                    "    public MemberInjectorRegistry() {",
+                    "      factories.put(\"test.TestASimpleRegistry\", new test.TestASimpleRegistry$$MemberInjector());",
```
stephanenicolas: This is a concern: it means that we instantiate all the injectors or the factories at startup when the registry itself is instantiated. This is probably an issue: it's gonna be very costly to use registries this way and startup time is probably gonna get much slower because of that.

stephanenicolas:
@pshmakov thx for this PR and the really good explanation you gave for this bug. The problem is clearly exposed.

But, we have a concern with this solution: registries are a minor optimization to prevent using reflection to find `Factories` and `MemberInjectors`. Your PR is gonna create a a overhead at startup of applications as it will trigger the instantiation of all member injectors and factories right at startup. As soon as registries are created, and this happens during the app's `onCreate` method, it will create possibly thousands of objects that will not only take RAM but also slow down startup.

Really the best if you really want to obfuscate your app is not to use registries at all. Did you do any benchmarking of your solution, it seems to us that it can only accomodate a small number of apps using DI only for a few classes, but for larger ones, the penalty will be quite big.

stephanenicolas:
Also, @pshmakov , @dlemures found a way to compromise and unify the current registry implementation and your proposal, that would also solve the issue: 

image


The idea is that
1) at the beginning of the registry, we have a map that maps class names to the buckect where a class is assigned.  This will later be proguarded and the bucket distribution will spread evenly the obfuscted classes.
2) instead of using the classname.hashcode to select the right bucketing method, we now use the map get the name of the bucketing method
3) bucketing methods themselves are unchanged.

We believe that this will create registries that can accomodate obfuscation.
Are you interested in implementing it ? It would be very appreciated as a contribution to TP.

Actually, if you would like to contribute even more, such a system would work without using hashcode at all. We could just slice the registries in groups of fixed size instead of bucketting them. The first (in order, and they are ordered since TP 1.1.2) 100 would go to bucketting method 1, the 101-200 would go to method 2, etc, etc.

pshmakov:
@stephanenicolas 
It's no secret that my "obfuscation friendly" implementation introduces some overhead compared to default one. However, I think you are greatly exaggerating the scale of this overhead.
I experimented by filling a Map<String, Object> with 1000 objects in application's onCreate method, and tested on various devices. Only on old and crappy ones (4.x) did the time go over 10ms, typically it was below 5ms. In massive apps that would have dependency graphs with thousands of nodes, the start time is typically on the order of a second in my experience.
The memory overhead is even more insignificant: it's 2 orders of magnitude less than total memory usage of an **empty** Android application (as created by Android Studio): ~150kb vs ~5-10Mb.

I also tested using https://github.com/stephanenicolas/DIDemoApps, and didn't see any drop in performance at all. But there there are only ~150 classes in the graph and the code that generates the classes is not in the repo, so I couldn't tweak that number easily. But we might as well test with larger numbers to be sure.  

Now, concerning you suggestion about improvement: if I understood you correctly, you say that the bucketing methods would still use a switch over class name, just as before. But that's exactly what doesn't work! I have a suspicion that you missed the part that not only does obfuscation break the bucketing mechanism, but it breaks the switches in the bucketing methods.

This code
`switch("com.smth.Smth") { case "com.smth.Smth": return true; } return false;`
returns false after obfuscation (if com.smth.Smth is a class that gets obfuscated).
That's because the compiler calculates the hashcode of "com.smth.Smth" in the line with "case", and basically converts the switch over strings into switch over ints. 
So, we can't have switches over class names.

But, if I understood your plan incorrectly, could you please show a sample code of a registry that implements it?

Also, as a bit of off-topic, I wonder why did you implement bucketing in the first place. I thought that the switch statement compiles to very efficient switch table bytecode, definitely with O(1) performance. And if such bucketing was indeed beneficial, I think it would have been implemented by java compiler devs long ago.

> We would like to understand if you gave a try to not using registries at all in your app.

It doesn't seem possible at all to marry the configuration with reflection and the obfuscation. The constraint factoryClassName = className +"$$Factory" is obviously broken by proguard. It seems, registries are our only hope :)

To conclude, my opinion is that the performance overhead is very minor, and it would better to merge this pr like it is. 
Keep in mind that
1. The alternative implementation is under the flag, so there's no detriment at all to the current version.
2. It's a blocker issue for introducing Toothpick to many projects (and probably gaining a lot of popularity in general). There are actually several teams waiting for this to be merged.

And if some ideas on how to reduce the performance overhead come around, they can be implemented later.

stephanenicolas:
@pshmakov we really want to have the registry configuration to be optimal. You might be right, I don't have the numbers but we don't want any slow down.

But it looks like there is a confusion here, our solution should work. I will detail right after answering your question: why bucketting ? 

The issue was that we have a large module in our application and without bucketting it would exceed the maximal length of a method for the java compiler....So yes, we need bucketting, or at least slicing.

Now, back to our proposal: you're right, we had saved the first swtich, but had an issue with the second one inside the bucketting methods. 

We took this very wise and (again) well explained problem into account and here is a new proposal:

At the begginning of the generated file we would do:

```java
mapFactoryToGroupIndex.put("Foo", 0);
mapFactoryToGroupIndex.put("Bar", 1);
...
mapFactoryToGroupIndex.put("Qurtz", 200);
...
```

Then, inside the lookup method we would have something like:
```java

int index = mapFactoryToGroupIndex.get(className);
//200 is the "group/bucket size", it can be fixed.
int indexInGroup = index % 200;
int groupIndex = index / 200;
switch(groupIndex) {
   case 1: return getFactory_1(indexInGroup);
   case 2: return getFactory_2(indexInGroup);
   ...
}
```

And each bucketting method can circumvent using the hashcode by doing:
```java
public <T> Factory<T> getFactory_1(int indexInGroup) {
  switch(indexInGroup) {
     case 0: return new Foo$$Factory();
     case 1: return new Bar$$Factory();
     ...
  }
}

public <T> Factory<T> getFactory_2(int indexInGroup) {
  switch(indexInGroup) {
     case 0: return new Qurtz$$Factory();
     ...
  }
}

```

pshmakov:
@stephanenicolas 
It seems you are right, I see that the initialization with your solution is several times faster than with mine on https://github.com/stephanenicolas/DIDemoApps
I guess it's because there's no classloading of each factory class while building a map.
Worth giving it a try :)

Btw, I believe `mapFactoryToGroupIndex` is actually `mapFactoryToIndex`

stephanenicolas:
Do you wanna implement it @pshmakov ? Thx for all this follow up already !

pshmakov:
I will implement it, no problem

stephanenicolas:
Thx a lot. Class loading is native on Android, and yes it consumes some time but it's not that bad. But instanciation is really slow.

pshmakov:
@stephanenicolas 
All done!

> Also, what's your full name to add you to the changelog of TP ?

Pavel Shmakov

stephanenicolas:
Thx @pshmakov . 

@dlemures , what do you think. IMHO, we could only keep the new implementation. There is no overhead and the code is simpler, the generated code too, and I prefer to avoid adding a new absconse flag and docs...

If you agree, @pshmakov , sorry but would you mind to clean up the PR and only keep the new implementation ? (Wait for the answer of @dlemures please).

dlemures:
Nice one!

I agree with @stephanenicolas, there is no point on keeping the other one.

pshmakov:
> If you agree, @pshmakov , sorry but would you mind to clean up the PR and only keep the new implementation ?

No problem, will do today.

pshmakov:
@stephanenicolas 
Removed the old implementation.
But there seems to be a small issue left: we also have to slice up the constructor code where the map is filled, for the same reason as the getter - method size limit.

pshmakov:
Fixed.
Also, I removed `int groupIndex = index / 200;` and just pass the  `index`. It's a slight simplification and also makes filling methods and getter methods beautifully symmetrical :)

toothpick-compiler/src/main/java/toothpick/compiler/registry/generators/RegistryGenerator.java
```diff
@@ -2,159 +2,215 @@
...
+  private final List<String> classNameList = new ArrayList<>();
 
   public RegistryGenerator(RegistryInjectionTarget registryInjectionTarget, Types types) {
     super(types);
     this.registryInjectionTarget = registryInjectionTarget;
+    factoryType = ParameterizedTypeName.get(ClassName.get(registryInjectionTarget.type), typeVariable);
+    clazzArgType = ParameterizedTypeName.get(ClassName.get(Class.class), typeVariable);
+
+    List<TypeElement> targetList = registryInjectionTarget.injectionTargetList;
+    for (int i = 0; i < targetList.size(); i++) {
```
stephanenicolas: we could have a for loop

toothpick-compiler/src/main/java/toothpick/compiler/registry/generators/RegistryGenerator.java
```diff
@@ -2,159 +2,215 @@
...
-    CodeBlock.Builder switchBlockBuilder = CodeBlock.builder().beginControlFlow("switch(bucket)");
-
-    List<MethodSpec> getterMethodForBucketList = new ArrayList<>(numOfBuckets);
-    Map<Integer, List<TypeElement>> getterMethodBuckets = getGetterMethodBuckets(registryInjectionTarget.injectionTargetList);
-    for (int i = 0; i < numOfBuckets; i++) {
-      List<TypeElement> methodBucket = getterMethodBuckets.get(i);
-      if (methodBucket == null) {
-        methodBucket = Collections.emptyList();
+  private void emitRegisterMethods(TypeSpec.Builder classBuilder) {
+    for (int i = 0; i < numGroups; i++) {
```
stephanenicolas: i --> groupIndex

toothpick-compiler/src/main/java/toothpick/compiler/registry/generators/RegistryGenerator.java
```diff
@@ -2,159 +2,215 @@
...
+    MethodSpec.Builder constructorBuilder = MethodSpec.constructorBuilder().addModifiers(Modifier.PUBLIC);
 
-    CodeBlock.Builder iterateChildAddRegistryBlock = CodeBlock.builder();
     for (String childPackageName : registryInjectionTarget.childrenRegistryPackageNameList) {
       ClassName registryClassName = ClassName.get(childPackageName, registryInjectionTarget.registryName);
-      iterateChildAddRegistryBlock.addStatement("addChildRegistry(new $L())", registryClassName);
+      constructorBuilder.addStatement("addChildRegistry(new $L())", registryClassName);
+    }
+
+    for (int i = 0; i < numGroups; i++) {
```
stephanenicolas: i -> groupIndex

toothpick-compiler/src/main/java/toothpick/compiler/registry/generators/RegistryGenerator.java
stephanenicolas: i -> factoryIndex

toothpick-compiler/src/main/java/toothpick/compiler/registry/generators/RegistryGenerator.java
```diff
@@ -2,159 +2,215 @@
...
-      switchBlockBuilder.addStatement("return ($L<T>) new $L$$$$$L()", typeSimpleName, getGeneratedFQNClassName(injectionTarget), typeSimpleName);
+    CodeBlock.Builder blockBuilder = CodeBlock.builder().beginControlFlow("if (index == null)");
+    blockBuilder.addStatement("return null");
+    blockBuilder.endControlFlow();
+    methodBuilder.addCode(blockBuilder.build());
+
+    methodBuilder.addStatement("int groupIndex = index / $L", groupSize);
+
+    CodeBlock.Builder switchBuilder = CodeBlock.builder().beginControlFlow("switch(groupIndex)");
+    for (int i = 0; i < numGroups; i++) {
```
stephanenicolas: i -> groupIndex

toothpick-compiler/src/main/java/toothpick/compiler/registry/generators/RegistryGenerator.java
```diff
@@ -2,159 +2,215 @@
...
-    getMethod.addCode(switchBlockBuilder.build());
-    return getMethod.build();
+    classBuilder.addMethod(methodBuilder.build());
   }
 
-  private Map<Integer, List<TypeElement>> getGetterMethodBuckets(List<TypeElement> injectionTargetList) {
-    int numOfBuckets = getNumberOfBuckets(injectionTargetList);
-    Map<Integer, List<TypeElement>> getterMethodBuckets = new HashMap<>();
+  private void emitGetFromGroupMethods(TypeSpec.Builder classBuilder) {
+    for (int i = 0; i < numGroups; i++) {
```
stephanenicolas: i -> groupIndex

toothpick-compiler/src/test/java/toothpick/compiler/registry/generators/RegistryGeneratorGroupSizeRule.java
```diff
@@ -0,0 +1,32 @@
+package toothpick.compiler.registry.generators;
+
+import org.junit.rules.TestRule;
+import org.junit.runner.Description;
+import org.junit.runners.model.Statement;
+
+public final class RegistryGeneratorGroupSizeRule implements TestRule {
```
stephanenicolas: not sure we really need this, but ok :)

stephanenicolas:
@pshmakov very nice implementation and thanks for slicing the map filling as well. I have reviewed the code and made minor comments. Thanks for the tests you added. 

I am trying a snapshot on our production app, with a benchmark test run. I will come back with the results here. If all is good and you can resolve the small comments I have, I think we should be good to release today.

toothpick-compiler/src/main/java/toothpick/compiler/common/ToothpickProcessor.java
```diff
@@ -591,4 +593,8 @@ private TypeElement getKindParameter(Element element) {
     TypeMirror firstParameterTypeMirror = ((DeclaredType) elementTypeMirror).getTypeArguments().get(0);
     return (TypeElement) typeUtils.asElement(typeUtils.erasure(firstParameterTypeMirror));
   }
+
+  protected CodeGenerator createRegistryGenerator(RegistryInjectionTarget target) {
+    return new RegistryGenerator(target, typeUtils);
```
dlemures: should we inline the method now?

pshmakov: If you want. I let it stay because it's called from 2 places, a very minor code duplication, but don't really care either way.

dlemures:
RTM

toothpick-compiler/src/main/java/toothpick/compiler/common/ToothpickProcessor.java
dlemures: yeah, but in the end it is just one line. So we are adding more lines than the ones we are removing/saving.

lukaville: it's not only about line count :) if we decide to change the body we can do it only in one place

dlemures: of course it is not the number of lines. But we are saving a line where we are just "newing" a class. If we changed the constructor, we would need to change the method probably as well. And moreover, we are not saving lines just making the code navigation more complex. 

So i don't see the benefit, as we are duplicating the method call anyway.

toothpick-compiler/src/test/java/toothpick/compiler/factory/FactoryRegistryTest.java
```diff
@@ -2,254 +2,459 @@
...
+                "      case 1: return (Factory<T>) new test.Class1$$Factory();", //
+                "      case 2: return (Factory<T>) new test.Class2$$Factory();", //
+                "    }", //
+                "    return null;", //
+                "  }", //
+                "", //
+                "  private <T> Factory<T> getFromGroup1(int index) {", //
+                "    switch(index) {", //
+                "      case 3: return (Factory<T>) new test.Class3$$Factory();", //
+                "      case 4: return (Factory<T>) new test.Class4$$Factory();", //
```
dlemures: wouldn't it be clearer to start the indexes from 0? Index within the Group.

toothpick-compiler/src/main/java/toothpick/compiler/common/ToothpickProcessor.java
pshmakov: Inlined

toothpick-compiler/src/test/java/toothpick/compiler/factory/FactoryRegistryTest.java
pshmakov: I think it's clearer this way: the indices here are exactly the same as in the corresponding map-filling method. And no need for a modulus operation.

dlemures: sounds good

dlemures:
Waiting for benchmark results

stephanenicolas:
Performance benchmark was long to run, but all is good on this side.

dlemures:
Merged! Thx @pshmakov

stephanenicolas:
@pshmakov : TP 1.1.3 is released and includes your contribution. It will be available on maven central within 3 hours.

Are you interested in becoming an official contributor to TP ? It basically boils down to creating PR on the main repo, being involved in triaging PRs, answering questions and development. We have a draft for TP 2 spec, and we hope to start developing it in 1-2 months.

pshmakov:
@stephanenicolas 
> Are you interested in becoming an official contributor to TP ?
Sure, I'm in.

pshmakov:
@stephanenicolas 
Btw, don't forget to update https://github.com/stephanenicolas/toothpick/wiki/Proguard-&-TP

stephanenicolas:
You've been invited, and I updated the wiki. Welcome to TP team !

pshmakov:
Yay!

> Since TP 1.1.3, TP now fully supports obfuscation, even when using registries.

That's not exactly true. Only with using registries, plus "adaptclassstrings" is required in proguard config.

stephanenicolas:
Thx for the update on the wiki @pshmakov

pshmakov:
@stephanenicolas 
> If you don't use registries, previous versions of TP will work well with obfuscation. You will need to preserve annotated classes.

Wait, what do you mean will work well? "You will need to preserve annotated classes." means "doesn't work" imo. And this was true for both configurations before 1.1.3. Or am I missing something?

stephanenicolas:
Preserving class/fields names is not "breaking" all of obfuscation. The code inside methods can still be ofuscated for instance. -donotobfuscate is not the same as keeping classes here and there.

pshmakov:
@stephanenicolas 
Well, I don't think anyone doubts that TP is compatible with obfuscation of stuff that is unrelated to DI, so no point stressing it.
Anyway, I think the wording now is a bit confusing. It may seem that we have broken something in 1.1.3, because it "worked very well" before, and now requires some steps. Plus, one may interpret it as if in previous versions the two configurations worked differently with respect to obfuscation, which isn't true afaik.
I suggest adding to the first sentence "(including obfuscation of names of annotated classes)", and then change the last paragraph to "If you are using configuration with reflection or older versions of TP, you need to keep the names of annotated classes".

pdapnz:
@pshmakov Is any idea why obfuscation in toothpick1.1.3 can be broken after gradle upgrade from `classpath 'com.android.tools.build:gradle:4.2.2'` to `classpath com.android.tools.build:gradle:7.0.1'` in new Android Studio Arctic Fox | 2020.3.1 Patch 1? 
NoFactoryFoundException exceptions start happening. When downgrade to gradle:4.2.2 all become fine again.

```
    toothpick.registries.NoFactoryFoundException: No factory could be found for class k9.e. Check that the class has either a @Inject annotated constructor or contains @Inject annotated members. If using Registries, check that they are properly setup with annotation processor arguments.
        at sc.b.c(FactoryRegistryLocator.java:3)
        at rc.f.b(ReflectionOffConfiguration.java:1)
        at rc.a.b(Configuration.java:1)
        at sc.b.a(FactoryRegistryLocator.java:1)
        at pc.d.a(InternalProviderImpl.java:10)
        at pc.i.a(ScopedProviderImpl.java:1)
        at pc.g.q(ScopeImpl.java:3)
        at pc.g.a(ScopeImpl.java:1)
```

PR #288, opened 2018-03-13
Added talks about Toothpick in Russian

terrakok:


stephanenicolas:
@terrakok so sorry for the long delay, and thank you very much for the talks and reporting them !

PR #317, opened 2019-01-06
Add "Scope.hasInstalledModules()" method for checking scope initialization.

terrakok:


terrakok:
@stephanenicolas hello! Can you review it? :smile:

PR #318, opened 2019-01-14
Add Toothpick.hasScope method for check scope in scopes tree.

terrakok:
Example of usage:
We can use it method for first initialization UI scope after Activity restore on Android OS.

At the moment we should to do like this:
[BaseFragment.kt](https://gitlab.com/terrakok/gitlab-client/blob/develop/app/src/main/java/ru/terrakok/gitlabclient/ui/global/BaseFragment.kt#L46)
😭😭😭

### DIFF
--- a/toothpick-runtime/src/main/java/toothpick/Toothpick.java
+++ b/toothpick-runtime/src/main/java/toothpick/Toothpick.java
@@ -30,15 +30,14 @@ public final class Toothpick {
 
   /**
    * Find scope by this {@code name} in current scopes tree.
-   * @return true if scope is exist.
+   * @return true if scope exist.
    */
   public static boolean hasScope(Object name) {
     if (name == null) {
       throw new IllegalArgumentException("null scope names are not allowed.");
     }
 
-    Scope scope = MAP_KEY_TO_SCOPE.get(name);
-    return scope != null;
+    return MAP_KEY_TO_SCOPE.containsKey(name);
   }
 
   /**
--- a/toothpick-runtime/src/main/java/toothpick/Toothpick.java
+++ b/toothpick-runtime/src/main/java/toothpick/Toothpick.java
@@ -29,8 +29,10 @@ public final class Toothpick {
   }
 
   /**
-   * Find scope by this {@code name} in current scopes tree.
-   * @return true if scope exist.
+   * Indicates whether a scope is open.
+   *
+   * @param name the name of the scope.
+   * @return true if the scope has been opened and not yet closed.
    */
   public static boolean isScopeOpen(Object name) {
     if (name == null) {
--- a/toothpick-runtime/src/main/java/toothpick/Toothpick.java
+++ b/toothpick-runtime/src/main/java/toothpick/Toothpick.java
@@ -29,6 +29,19 @@ public final class Toothpick {
   }
 
   /**
+   * Find scope by this {@code name} in current scopes tree.
+   * @return true if scope is exist.
+   */
+  public static boolean hasScope(Object name) {
+    if (name == null) {
+      throw new IllegalArgumentException("null scope names are not allowed.");
+    }
+
+    Scope scope = MAP_KEY_TO_SCOPE.get(name);
+    return scope != null;
+  }
+
+  /**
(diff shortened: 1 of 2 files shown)


rovkinmax:
Nice! I've been waiting for this feature for a long time

toothpick-runtime/src/main/java/toothpick/Toothpick.java
```diff
@@ -28,6 +28,19 @@ private Toothpick() {
     throw new RuntimeException("Constructor can't be invoked even via reflection.");
   }
 
+  /**
+   * Find scope by this {@code name} in current scopes tree.
+   * @return true if scope is exist.
```
dimsuz: "if scope exists" would be more grammatically correct here

toothpick-runtime/src/main/java/toothpick/Toothpick.java
```diff
@@ -28,6 +28,19 @@ private Toothpick() {
...
+   */
+  public static boolean hasScope(Object name) {
+    if (name == null) {
+      throw new IllegalArgumentException("null scope names are not allowed.");
+    }
+
+    Scope scope = MAP_KEY_TO_SCOPE.get(name);
+    return scope != null;
```
dimsuz: why not `MAP_KEY_TO_SCOPE.contains(name)`? It seems like `get` would be needed in cases where it is possible to have a `name` mapped to a `null` scope object. Is this possible in toothpick? (I don't know internals enough to answer that)

terrakok: Hmm. May be you're right. But scope can't be null for Toothpick.

terrakok:
@stephanenicolas @dlemures what do you think about it?

toothpick-runtime/src/main/java/toothpick/Toothpick.java
```diff
@@ -28,6 +28,18 @@ private Toothpick() {
     throw new RuntimeException("Constructor can't be invoked even via reflection.");
   }
 
+  /**
+   * Find scope by this {@code name} in current scopes tree.
+   * @return true if scope exist.
+   */
+  public static boolean hasScope(Object name) {
```
stephanenicolas: can you rename to `isScopeOpen` please ?

And please add unit tests to : `toothpick.ToothpickTest`. We will release it as part of TP 2.1. Thx @terrakok

terrakok:
@stephanenicolas  done :ok_hand:

toothpick-runtime/src/test/java/toothpick/ToothpickTest.java
```diff
@@ -236,6 +236,55 @@ public void reset_shouldCallResetForProvidedScope() throws Exception {
     verify(mockScope);
   }
 
+  @Test
+  public void isScopeOpen_shouldReturnFalse_WhenThisScopesWasNotCreated() throws Exception {
+    //GIVEN
+
+    //WHEN
+    Boolean isFooScopeOpen = Toothpick.isScopeOpen("foo");
```
stephanenicolas: please use a small b boolean.

toothpick-runtime/src/main/java/toothpick/Toothpick.java
```diff
@@ -28,6 +28,18 @@ private Toothpick() {
     throw new RuntimeException("Constructor can't be invoked even via reflection.");
   }
 
+  /**
+   * Find scope by this {@code name} in current scopes tree.
+   * @return true if scope exist.
```
stephanenicolas:
```java
/**
Indicates whether a scope is open.
@return true iff the scopehas been opened and not yet closed.
*/
```

stephanenicolas:
We will be glad to merge and release after the comments are addressed. Thx @terrakok

toothpick-runtime/src/test/java/toothpick/ToothpickTest.java
terrakok: Oops :) It's Kotlins habit

dlemures:
Looks great :)

Issue #316, opened 2018-12-04
KTP draft specs


dlemures:
Kotlin Toothpick draft specs - **WIP**

### Use delegate properties for field injection
We should not use annotations for field injection, instead we will use delegate properties. For example:
```java
by ktp.inject() // default scope
by ktp.lazy() // lazy
by ktp.provider() // provider
```

Moreover, there should be a way to:
* Specify when to use the default scope and when to wait for a scope to be provided.
* Use a named injection.

### Kotlin friendly API for
* Module and bindings definition (https://github.com/sporttotal-tv/toothpick-kotlin-extensions and https://github.com/stephanenicolas/toothpick/issues/305)
* Scope creation and "scope binding".
* Configuration of injectable classes
  * Should we use the `@Inject` annotation? and where?
  * How to define the constructor to use
  * How to define a singleton
  * How to define in what scope an instance will be created

### Generate code compatible with optional parameters (possible?)

WIP

Moreover, KTP will be based on TP 2.0 -> no registries (no configuration needed), incremental annotation processing, better scope annotations, ... (https://github.com/stephanenicolas/toothpick/issues/315, https://github.com/stephanenicolas/toothpick/issues/225)

*Something you miss? Any suggestion? Please comment 🙂*

code-twister:
Can we have something like this for named injections:
```
by ktp.named("THE_NAME").inject() // default scope
```

code-twister:
I think I can answer some of you questions @PaulWoitaschek:

1. KTP comes from: **K**otlin **T**ooth**P**ick
2. Based on the initial KTP proposal the class you provided would look something like:
```
class MyInjectable : MyFrameworkClass(){

  val myNamedClass : Sth by ktp.named("whatever").inject()
  val sthElse : SthElse by ktp.inject()

  ...
}
```

Note the lack of **lateinit** or the annotations. Another small change but with a big impact: you can use val instead of var. I think it's a good idea to go with delegated properties (https://kotlinlang.org/docs/reference/delegated-properties.html). The code looks much cleaner.

I think scopes require further investigation on how they can be achieved in pure kotlin and with the new API.

PaulWoitaschek:
With comes from I don't mean where the name comes from but rather: where does the reference come from?

If it's a top-level property we can use this only without scopes.
I don't currently see a way of archieving this without the annotation.

GisoBartels:
For scopes, what do you think about this idea?

```Kotlin
interface ToothpickScope {
[...]

Issue #305, opened 2018-10-23
Kotlin optimizations


markchristopherng:
Hi,

Will you guys be adding any Kotlin extensions in the future?

PaulWoitaschek:
There are some things that would be nice for kotlins.
* The `bind(x).to(y)` syntax is bad as it conflicts with kotlins `to` extension function.
It suggest adding a `bindToClass` method alias because that's more convenient on kotlin
* All that bind-to-withProvider-singletonInInstance could be abstracted into a convenient DSL.
* We could have an extension function on with the scope as the receiver for simpler injection without that Toothpick.inject boilerplate:
```kotlin
fun Scope.inject(target: Any) {
  Toothpick.inject(target, this)
}
```

chenthorne:
We've started using a set of custom written extensions that we'd like to make public and incorporate with the toothpick library as a new module. There's a sample project at https://github.com/Leafly-com/toothpick-extensions-sample that has the extension code as well as examples on how it could be used.

- Extensions to allow use of Kotlin class objects (no more `.java`)
- Simple set of infix functions with a DSL for declaring `Module`s
- Kotlin Delegate based lazy injection that allows `private val dep: Dep by inject()`
- Android specific extensions that search for a scope or assume a default one for activities and fragments
- Inject with late created scope but still specify as private class member

Quick Preview:
```kt
// Inject with Kotlin Class
scope.getInstance(KotlinClass::class)

// In a Module, bind with Kotlin Class
bind(KotlinClass::class)...

// Inject dependency safely without specifying type via reified types
val dependency: Dependency = inject(scope)

// Lazy inject dependency via delegate
val dependency: Dependency by injectLazy(scope)

// Will inject with Toothpick.openScope(activityRef.application, activityRef)
class MyActivity : Activity() {
    private val dependency1: Dependency by inject()
}

// Will inject with lazy provided scope
class MyActivity : Activity() {
    private lateinit var lateScope: Scope
    private val dependency1: Dependency by inject { scope = lateScope }
    ...
}

// Binding functions
bind(Dep::class) providedBy DepProvider()
bind(Dep::class) singletonPovidedBy DepProvider::class
bind(Dep::class) named "DepName" providedBy DepProvider()

// DSL
module {
   Picasso::class providedBy PicassoProvider::class 
}
```

dimsuz:
We have a shorter `getInstance()`, not based on a `KClass` like yours though, works great for us too:

```kotlin
[...]

PR #324, opened 2019-02-12
Initial Kotlin Extensions for Toothpick

chenthorne:
Adds two new modules, `toothpick-kotlin` and `toothpick-kotlin-androidx`, to provide kotlin friendly interfaces to common Toothpick use cases. Also provides some syntactic sugar for using Toothpick with Android Kotlin.

Both modules contain README's with examples on how to use the extensions and the `smoothie-sample` project has been updated to have a mirror of the existing samples but in kotlin.

See issues #305 and #316 or the draft extensions at https://github.com/Leafly-com/toothpick-extensions-sample

Things remaining:
- [ ] Get community and project maintainer buy-in
- [ ] Add tests for both modules
- [ ] Determine if kotlin samples should be split off into their own module
- [ ] Make sure CI is happy and checking


chenthorne:
@code-twister with modules split

smoothie-sample/src/main/java/com/example/smoothie/kotlin/LessSimpleActivity.kt
```diff
@@ -0,0 +1,49 @@
...
+    private val title by lazy { findViewById<TextView>(R.id.title) }
+    private val subTitle by lazy { findViewById<TextView>(R.id.subtitle) }
+
+    override fun onCreate(savedInstanceState: Bundle?) {
+        val scope = Toothpick.openScopes(application, this)
+        scope.installModules(SmoothieActivityModule(this))
+        scope.installModules(module { String::class named "SpecialString" providedBy "My Special String" })
+        super.onCreate(savedInstanceState)
+        setContentView(R.layout.simple_activity)
+        ButterKnife.bind(this)
```
stephanenicolas: Do we still need butterknife ? It seems to me we're using a simple lazy to assign the views.

toothpick-kotlin/src/main/java/toothpick/kotlin/ToothpickInjectDelegate.kt
```diff
@@ -0,0 +1,45 @@
...
+    override fun getValue(thisRef: OWNER, property: KProperty<*>): T {
+        return value ?: injectValue(thisRef, property)
+    }
+
+    protected open fun injectValue(thisRef: OWNER, property: KProperty<*>): T {
+        val injectConfig = InjectConfig()
+        if (injectConfigBlock != null) {
+            injectConfig.apply(injectConfigBlock)
+        }
+        val scope = injectConfig.scope ?: scope ?: getScope(thisRef, property) ?: throw IllegalStateException("No valid scope available when attempting to inject $property in $thisRef")
```
stephanenicolas: Let's finish with a `.` and explain how to fix the issue. If there is more than one alternative, let's explain them all.

toothpick-kotlin/src/main/java/toothpick/kotlin/ToothpickInjectDelegate.kt
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    protected open fun getScope(thisRef: OWNER, property: KProperty<*>): Scope? {
+        if (scope != null) {
+            return scope
+        }
+
+        if (thisRef is HasScope) {
+            return thisRef.scope
+        }
```
stephanenicolas: I got an issue with this. No idea if it is a standard way to implement other DI engines usually. 
But it seems very hacky for a few reasons. I understand that we have this mechanism, and the fallback of creating a new scope right under the app for activities, because we want to be more concise and not pass the scope when doing `by inject(scope)`

But first things first, I am not sure I understand how lazy and inject differ now. Isn't the delegated property going to be "computed" only when the code accesses it ? if so, what's the difference with a lazy injection, and if not, when is it gonna be computed?

I don't really like the interface, and the fallback neither. It assumes that the activity as a given property, implements a given interface. An activity could decide to call it's scope `scopeA` and forget about the interface and no injection would take place.
Same for the fallback, we don't want to consider the activity's scope is placed right under the application by default. And for fragments, I am not even sure that the activity scope should be the default.

I was thinking that it would be maybe cleaner:
1) to pass the scope all the time when injecting / lazying a property. `by inject(scope)`. We could also prefer to use a scope name `by inject(this)`
2) the activity can itself have a getter/delegate to create it's scope when needed
```kotlin
  lateinit var scope: Scope
     get() { 
            TP.createScope(...)
            scope.installModules(...)
            return scope
    }
```
3) if it's not convenient to pass the scope to every inject call. Can't we have another mechanism where we would restor the call to TP.inject(this, scope) and this would provide the scope to `by inject()` 's delegate?

toothpick-kotlin/src/main/java/toothpick/kotlin/ToothpickExtensions.kt
```diff
@@ -0,0 +1,52 @@
...
+
+fun <T : Any> Module.bind(clazz: KClass<T>): Binding<T> {
+    return bind(clazz.java)
+}
+
+fun Scope.inject(toInject: Any) {
+    Toothpick.inject(toInject, this)
+}
+
+fun <T : Any> Scope.getInstance(clazz: KClass<T>, name: String? = null, annotationName: KClass<out Annotation>? = null): T {
```
stephanenicolas: do we need the non reified ones ?

toothpick-kotlin/src/main/java/toothpick/kotlin/ModuleDSL.kt
```diff
@@ -0,0 +1,76 @@
...
+
+    inline infix fun <reified T : Any> KClass<T>.named(name: KClass<out Annotation>): Binding<T> {
+        return module.bind(T::class.java) named name
+    }
+
+    inline infix fun <reified T : Any> KClass<T>.named(name: String): Binding<T> {
+        return module.bind(T::class.java) named name
+    }
+
+    inline infix fun <reified T : Any> KClass<T>.providedAs(clazz: KClass<out T>): Binding<T>.BoundStateForClassBinding {
```
stephanenicolas: all the fluent API has been renamed entirely. Can we keep the exact same method names for the bindings ?
I am not opposed to renaming them if needed, but it's a big change for TP. Are there any kotlin "way of doing things" that are at stake with these new names ?
Oh.. `to` is a kotlin keyword?

stephanenicolas: I think the DSL should look more like:
```kotlin
bind T::class.java toInstance t0
bind T::class.java toClass T1::class.java
bind T::class.java withName "foo" toInstance t0
bind T::class.java toProviderInstance p0
bind T::class.java toProviderClass P::class.java

for `providesSingletonInScope`, `instancesInScope`, `singletonInScope`, I agree we have an issue that a infix method needs a param...
```

toothpick-kotlin/README.md
```diff
@@ -0,0 +1,115 @@
...
+
+// Inject with Kotlin Class
+scope.getInstance(KotlinClass::class)
+// Inject with reified type
+val type: KotlinClass = scope.getInstance()
+
+// In a Module, bind with Kotlin Class
+module.bind(KotlinClass::class)...
+
+// Inject dependency without specifying type
```
stephanenicolas: the type ?

toothpick-kotlin/README.md
```diff
@@ -0,0 +1,115 @@
...
+scope = Toothpick.openScope(...)
+lateName = "myLateName"
+dependency.foo()           // Injection happens here
+namedDependency.foo()      // Injection happens here
+lateNamedDependency.foo()  // Injection happens here
+
+// Lazy inject with scope search
+class ScopeSearch : HasScope {
+    override lateinit var scope: Scope
+        protected set
```
stephanenicolas: I would prefer an open get() that subclasses can override than a set.

stephanenicolas:
One side effect of this PR is that member injectors are not needed for kotlin code.

toothpick-kotlin/src/main/java/toothpick/kotlin/ToothpickInjectDelegate.kt
stephanenicolas: And there is another issue. IMHO. We're loosing the `@Inject` annotation for fields. It seems to me there is a tension in your PR between what happened to butterknife (replaced by lateinit `by findViewByID`) and the classic `@Inject` annnotation. I prefer the annotation, but why did you prefer the  other way, for speed ?

stephanenicolas:
@chenthorne can you have a look at kodeint DSL for bindings / module. They use the method with elegantly : https://github.com/Kodein-Framework/Kodein-DI

smoothie-sample/src/main/java/com/example/smoothie/kotlin/LessSimpleActivity.kt
chenthorne: Not really, was just copying the Java examples over for parity.

toothpick-kotlin/src/main/java/toothpick/kotlin/ToothpickInjectDelegate.kt
chenthorne: I'm not a fan of this hack. I haven't found a way to do this better without passing scope with every call. At one point, I had scope as nullable, so if you didn't provide it the delegate would try to use `this`. I didn't really like that either but it may be better than this route.

In most cases, I don't want to have to pass scope ever unless I'm doing something fancy.

I like no longer having the `@Inject` annotations, less non-code in my code to read around. We gain a few things with it:
1. No longer need an annotation processor, quicker build time
2. No longer need member injectors, quicker build time, no java leaking into pure Kotlin modules causing more compilation
3. Private injected members, no need to weaken access modifier to allow an APT generated class access. 
4. No longer need to set as `lateinit var` or nullable
5. IMO, reads cleaner
6. Puts all injection logic in one spot as you can put scope and name in there as well, it's all done in one go instead of multiple annotations or split of logic.

toothpick-kotlin/src/main/java/toothpick/kotlin/ToothpickExtensions.kt
chenthorne: When you want to inject a subclass into a super.

toothpick-kotlin/src/main/java/toothpick/kotlin/ModuleDSL.kt
chenthorne: Yeah, I went back and forth on this one a lot. Poked around in kodein and other kotlin DSLs to find a common naming scheme and didn't really see anything definitive. I settled on what I did as it read the most naturally to me as a human, but I'm not hugely attached to it. I like the examples you've provided.

toothpick-kotlin/README.md
chenthorne: Without passing the class (e.g., `Dependency::class`), type is the wrong word here.

toothpick-kotlin/README.md
chenthorne: I'm not sure I follow the comment. I settled with this as it provided universal/public access to the getter but only protected access to the setter. This keeps any ole class changing the scope.

chenthorne:
Thanks for the review!

@stephanenicolas Sure thing, I'll look through that again, anything in Kodein you'd like me to look at in particular? Did you look at Koin? Some of the stuff I did was driven by things I liked in there.

Looking forward to chatting with you about this soon.

stephanenicolas:
Yeah I like the binding DSL in kodein.
bind Foo::class.java with singleton { lambda }

toothpick-kotlin/src/main/java/toothpick/kotlin/ToothpickInjectDelegate.kt
stephanenicolas: 1. we still need the factories to be created.
2. factories could be implemented in kotlin using extension function but I don't know any AP doing this. Not sure the build times would be any better.
3. that one ok
4. I am really motivated by keystroke count.`@Inject lateinit var foo` vs `lateinit var foo by inject()`
5...
6. It also changes the logic of the instantiation and scopes. Now the scope would be decided by the client whereas in java it's set by the entity itself. I feel that it might be conflicting in some cases too.

chenthorne: 1. Yup, fair point for some classes, others won't need it like activities and fragments (nice for Android land, not applicable for Java land).
2. Sure, factories are still needed for most, and there is still Java leakage, but less and less to get rid of if TP ever moves to a kotlin based solution.
3. A big plus in my eyes as it influences 4
4. Actual code:
`@Inject lateinit var foo: Foo`
`private val foo: Foo by inject()`

I really like the second vs the first, and in the day of IDEs and code complete, keystroke count alone is a poor argument loosening access restrictions and adding mutability to the variable. Big philosophy of Kotlin is immutability, and this change moves TP in that direction.

5. -
6. In my use cases, it hasn't been much of a big deal as I rarely am doing anything special with scopes. And when I am, I'm already paying extra attention to how I'm using the scope. I'm not sure I follow what you mean by `client` and `entity`, could you elaborate?

toothpick-kotlin/src/main/java/toothpick/kotlin/ToothpickExtensions.kt
stephanenicolas: Is there an example in your PR ?

toothpick-kotlin/src/main/java/toothpick/kotlin/ToothpickInjectDelegate.kt
stephanenicolas: 2. Do kotlin people really talk about java leakage ? I find it funny ;)

I get it for #4. That's a good point (sorry I am really new to kotlin, it takes some time to put the pieces together ;)). Yes for immutability, it's a big win. But I like my so much my little cute annotations ! Snifff...

6. I understand why the PR goes a little too far with scopes now.  ;)

toothpick-kotlin/src/main/java/toothpick/kotlin/ModuleDSL.kt
stephanenicolas: I also like the with style ... but really the main problem is that `to` is a reserved word.. I don't find it great I must say. 
Same goes for `as` , otherwise we could have had construct like
```
bind T::class.java to instance t0
bind T::class.java to instance t0 as singleton
```
This would be the more natural solution. 
Would it work if we would do
```
bind T::class.java `to` instance t0
bind T::class.java `to` instance t0 `as` singleton
```

It's ugly, but I am trying to see all our options.

smoothie-sample/build.gradle
```diff
@@ -70,3 +78,9 @@ dependencies {
   androidTestImplementation deps.easymock
   androidTestImplementation deps.dexmaker
 }
+
+tasks.withType(org.jetbrains.kotlin.gradle.tasks.KotlinCompile).all {
```
stephanenicolas: we should use Task Configuration Avoidance.

toothpick-kotlin/src/main/java/toothpick/kotlin/ModuleDSL.kt
stephanenicolas: We could actually allow the to operator if module would support adding pairs..
```
module {
   bind Class<T> to instance(t0)
   bind Class<T> to Class<T1> 
   bind Class<T> to providerInstance(p0)
   bind Class<T> to Class<Provider<P1>>
}
```

chenthorne: Interesting. I'll have to play with this to see if how it feels in real code. We're using a variant of the DSL internally so I can see how it goes with real world modules.

toothpick-kotlin/src/main/java/toothpick/kotlin/ToothpickExtensions.kt
chenthorne: divide into two overloads to disambiguate name or annotationName, not both

dlemures:
Closing. We already released KTP.

Thx :)

PR #341, opened 2019-07-22
Sni/tp 3 sample

stephanenicolas:
@dlemures can you check it please. 
I add:
* new ways to configure a scope when opening it. The new lambda is only applied on creation.
* new way to open a child scope 
* documentation for everything
* tests
* changing the sample accordingly, start working on a kotlin sample


dlemures:
RTM

PR #355, opened 2019-08-07
increase code coverage

stephanenicolas:
Increase code coverage.


toothpick-runtime/src/main/java/toothpick/configuration/MultipleRootScopeCheckOnConfiguration.java
```diff
@@ -26,21 +26,17 @@
...
     }
 
-    if (scope == rootScope) {
+    if (newRootScope == rootScope) {
       return;
     }
 
-    if (scope.getParentScope() != null) {
-      return;
-    }
```
dlemures: why do we remove this check?

stephanenicolas:
Because it was not possible to cause this method with a child that had a parent, there is no path in the code for that.

Issue #360, opened 2019-08-23
superMemberInjector generated with missing type information


zawadz88:
In Toothpick 3.0.0 there's an issue where the generated `__MemberInjector` has an incorrect `superMemberInjector` field i.e. it does not set the type argument correctly.

E.g. assume we have `AddNewActivity` with `@Inject` annotated fields. `AddNewActivity` extends `BaseActivity` which has a type parameter e.g.
```

public class AddNewActivity extends BaseActivity<Integer> {
 // ...
}
```
where `BaseActivity` is: 
```
abstract class BaseActivity<T> : SomeInterfaceWithGeneric<T>, AppCompatActivity() {

    @Inject
    lateinit var dummyDependency: DummyDependency
}
```

When you try to build the project KAPT will fail with the following reason:
```
> Task :kaptDebugKotlin FAILED
e: /Users/me/workspace/toothpick/toothpick-sample/build/generated/source/kapt/debug/com/example/toothpick/activity/AddNewActivity__MemberInjector.java:9: error: cannot find symbol
  private MemberInjector<BaseActivity<T>> superMemberInjector = new com.example.toothpick.activity.BaseActivity__MemberInjector();
                                      ^
  symbol:   class T
  location: class AddNewActivity__MemberInjector
e: /Users/me/workspace/toothpick/toothpick-sample/build/generated/source/kapt/debug/com/example/toothpick/activity/AdvancedBackpackItemsActivity__MemberInjector.java:9: error: cannot find symbol
  private MemberInjector<BaseActivity<T>> superMemberInjector = new com.example.toothpick.activity.BaseActivity__MemberInjector();
```
Generated `AddNewActivity__MemberInjector`

```
public final class AddNewActivity__MemberInjector implements MemberInjector<AddNewActivity> {
  private MemberInjector<BaseActivity<T>> superMemberInjector = new com.example.toothpick.activity.BaseActivity__MemberInjector();

  @Override
  public void inject(AddNewActivity target, Scope scope) {
    superMemberInjector.inject(target, scope);
    target.backpackItemValidator = scope.getInstance(BackpackItemValidator.class);
  }
}
```

I've added an example in `toothpick-sample` app to reproduce this:
https://github.com/stephanenicolas/toothpick/compare/master...zawadz88:super-member-injector-bug

[...]

PR #361, opened 2019-08-26
superMemberInjector with generic type fix

zawadz88:
Fixes issue #360 where the generated __MemberInjector has an incorrect superMemberInjector field i.e. it does not set the type argument correctly.


stephanenicolas:
When green, @dlemures will merge & release within 24 hours. I got.in touch
with him.

Thx Piotr

Issue #362, opened 2019-08-29
KTP Binding API does not work well with IDE auto-completion


dlemures:
When using the KTP binding API on AndroidStudio/IJ, the auto-completion does not work very well. The issue is that the extensions methods on top of Binding have the same name as methods already available on the same class. In order to import the right method, it is necessary to scroll and try every suggestion...

Before
![image](https://user-images.githubusercontent.com/5851365/63958539-39c92e80-ca93-11e9-9ac8-a80da1bb0943.png)

Only one of the ones above is valid. Some of the extension functions are missing.

After
![image](https://user-images.githubusercontent.com/5851365/63958793-857bd800-ca93-11e9-851c-6fd9742fb66c.png)

All the above are valid.

dlemures:
PR created https://github.com/stephanenicolas/toothpick/pull/363

dlemures:
PR merged

PR #363, opened 2019-08-29
Fixing issue 362 and improvement

dlemures:
This PR contains the following changes:

* Solving [issue 362](https://github.com/stephanenicolas/toothpick/issues/362). In order to fix the problem, it moves KTP binding API to use classes instead of extension functions.
* Cleaning KTP class (no need to have companion object or extend Toothpick) and moving injector replacement logic to different class (single responsibility). Also it will not make it possible to write (the suggested way by the IDE...):
```kotlin
KTP.TP.openScope...
```
* Trigger member injection (annotated fields and annotated) even when delegates are used. This will allow to use method injection on roots using KTP.

These changes are improvements that do not change the public API.

### DIFF
--- a/ktp/src/main/kotlin/toothpick/InjectorReplacer.kt
+++ b/ktp/src/main/kotlin/toothpick/InjectorReplacer.kt
@@ -1,3 +1,19 @@
+/*
+ * Copyright 2019 Stephane Nicolas
+ * Copyright 2019 Daniel Molinero Reguera
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *        http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
 package toothpick
 
 import toothpick.ktp.KTP
--- a/ktp/src/main/kotlin/toothpick/InjectorReplacer.kt
+++ b/ktp/src/main/kotlin/toothpick/InjectorReplacer.kt
@@ -9,12 +9,12 @@ object InjectorReplacer {
 
     fun replace() {
         Toothpick.injector = object : InjectorImpl() {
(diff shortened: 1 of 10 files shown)


ktp/src/main/kotlin/toothpick/InjectorReplacer.kt
```diff
@@ -0,0 +1,36 @@
...
+ * limitations under the License.
+ */
+package toothpick
+
+import toothpick.ktp.KTP
+
+/**
+ * Replaces Toothpick injector to be able to notify delegates.
+ */
+object InjectorReplacer {
```
code-twister: can you make this internal?

ktp/src/main/kotlin/toothpick/ktp/KTP.kt
```diff
@@ -25,32 +25,23 @@ import toothpick.ktp.delegate.DelegateNotifier
...
-                override fun <T : Any> inject(obj: T, scope: Scope) =
-                    if (delegateNotifier.hasDelegates(obj)) {
-                        delegateNotifier.notifyDelegates(obj, scope)
-                    } else {
-                        super.inject(obj, scope)
-                    }
-            }
-        }
+    init {
+        InjectorReplacer.replace()
```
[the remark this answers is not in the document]
dlemures: cannot do it cause of visiblitiy. Otherwise we would need to make the field protected (instead of package-private) and extend Toothpick (which means, having to go back to companion and `KTP.TP....`)

ktp/src/main/kotlin/toothpick/InjectorReplacer.kt
dlemures: sure

toothpick-runtime/src/main/java/toothpick/Toothpick.java
```diff
@@ -39,7 +39,7 @@
   // its transformation
   private static final ConcurrentHashMap<Object, Scope> MAP_KEY_TO_SCOPE =
       new ConcurrentHashMap<>();
-  protected static Injector injector = new InjectorImpl();
```
dlemures: better package-private

ktp/src/main/kotlin/toothpick/InjectorReplacer.kt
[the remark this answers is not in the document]
dlemures: renamed. 

I tried to use an extension functions from within the same package, but it doesnt work cause of the visibility. :(

stephanenicolas:
RTM. Thx a lot. Good changes ;)

PR #364, opened 2019-08-29
WIP default scope implementation, just a proof of concept first

code-twister:
Just getting this out there to see what people think.
I'm planning to add a wiki page for it.
I think it will help with on-boarding people using TP without scopes.

### DIFF
--- a/toothpick-runtime/src/main/java/toothpick/configuration/Configuration.java
+++ b/toothpick-runtime/src/main/java/toothpick/configuration/Configuration.java
@@ -30,6 +30,7 @@ public class Configuration
   private RuntimeCheckConfiguration runtimeCheckConfiguration = new RuntimeCheckOffConfiguration();
   private MultipleRootScopeCheckConfiguration multipleRootScopeCheckConfiguration =
       new MultipleRootScopeCheckOffConfiguration();
+  boolean usesScopes = true;
 
   /**
    * Performs many runtime checks. This configuration reduces performance. It should be used only
@@ -82,6 +83,11 @@ public class Configuration
     return this;
   }
 
+  public Configuration disableScopes() {
+    this.usesScopes = false;
+    return this;
+  }
+
   @Override
   public void checkIllegalBinding(Binding binding, Scope scope) {
     runtimeCheckConfiguration.checkIllegalBinding(binding, scope);
@@ -105,5 +111,9 @@ public class Configuration
   @Override
   public void onScopeForestReset() {
     multipleRootScopeCheckConfiguration.onScopeForestReset();
+  }
+
+  public boolean usesScopes() {
+    return usesScopes;
   }
 }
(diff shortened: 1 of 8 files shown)


toothpick-runtime/src/main/java/toothpick/configuration/Configuration.java
```diff
@@ -30,6 +30,7 @@
   private RuntimeCheckConfiguration runtimeCheckConfiguration = new RuntimeCheckOffConfiguration();
   private MultipleRootScopeCheckConfiguration multipleRootScopeCheckConfiguration =
       new MultipleRootScopeCheckOffConfiguration();
+  boolean usesScopes = true;
```
dlemures: This should be a configuration, similar to the multiple root config.

stephanenicolas:
@code-twister, @dlemures and I had discussed an alternative to this: 
Being able to use a rootScope directly. The root scope would be created on demand. If there is a root scope such an application scope, they would be equal. 

I would prefer this the join default scope. It would also provide a better separation of concern in the TP code. 

In your PR, can you also provide an example of the API you create, it would be easier to review and see how good and valuable the API is.

code-twister:
@stephanenicolas, @dlemures and I also talked about this. The main motivation is to have a simple API when you don't need scopes, so you don't need to know about the existence of scopes, that's why it's just `Toothpick.inject(obj)`. 

Another important aspect is to make sure people don't use both variants so it should fail if you're trying to use scopes when they are disabled and also fail when you try the scopeless API when scopes are enabled.

stephanenicolas:
I understand. To my mind the problema with this approach are:
1) it introduces a new concept not default scope. Root scope does not.
2) the seperation of concerns is better when we delegate to a scope. Here
you add the scope API to the Toothpick class. It is an issue for
maintenance.
3) as you said, a default scope introduces a mode that is not compatible
with the scope mode.
The idea with the root scope was to make it easy to deal with a single
scope, and still let users scale up when their app grows in complexity.

getRootScope() is at most 3 key strokes in IntelliJ/AS: 'grs'.

I prefer this approach. TP 3 is already a little bit more spaghetti than TP
2. Argument 2 above is a big one to reduce maintenance.

And by the way, the root scope stuff is interesting to code. Not so trivial
;)

S.

toothpick-runtime/src/main/java/toothpick/configuration/Configuration.java
code-twister: changed

code-twister:
@stephanenicolas I have to agree with all of your points, and I also understand your preference. The solution I propose is different in the sense that people don't have to know anything about scopes to use TP, they don't even have to know scopes are being used in the background. I guess I'm thinking a lot about going KISS here. Having a default root scope that you can extend will be a bit more confusing for inexperienced users (not to mention interesting/difficult to code).

For your second argument: I could move .inject() and .installModules() into a separate class. I thought it made more sense for the methods to be in Toothpick. I guess it all comes down to preference.

stephanenicolas:
@code-twister the PR is quite good and I agree, it's not that complicated. It's pretty clean and thanks for that. 

But what do you think of keeping it as is, and opening another one with the rootScope idea so that we can compare ? 
I feel that we might like the other better still for the reasons I mentioned above.

Thx for the contribution to TP, it's appreciated. And the quality is as good as anything I've seen from you in the past: excellent ;)

code-twister:
closing this in favour of https://github.com/stephanenicolas/toothpick/pull/365

PR #365, opened 2019-08-30
WIP rootScopes implementation prototype

code-twister:
Based on a discussion with @dlemures and @stephanenicolas this is an alternative to https://github.com/stephanenicolas/toothpick/pull/364

Just looking for feedback so far, still need to add docs and extend KTP.

### DIFF
--- a/toothpick-runtime/src/main/java/toothpick/Toothpick.java
+++ b/toothpick-runtime/src/main/java/toothpick/Toothpick.java
@@ -39,10 +39,35 @@ public class Toothpick {
   // its transformation
   private static final ConcurrentHashMap<Object, Scope> MAP_KEY_TO_SCOPE =
       new ConcurrentHashMap<>();
+  // ConcurrentHashSet doesn't exist so we're not using a Set ;)
+  private static final ConcurrentHashMap<Object, Scope> ROOT_SCOPES = new ConcurrentHashMap<>();
   static Injector injector = new InjectorImpl();
 
   protected Toothpick() {
     throw new RuntimeException("Constructor can't be invoked even via reflection.");
+  }
+
+  /**
+   * Returns or opens a root scope as follows:<br>
+   *
+   * <ul>
+   *   <li>if there's an existing scope tree the root scope of that tree;
+   *   <li>creates a default root scope if there's none;
+   *   <li>throws an exception if there are multiple scope trees (e.g. a forest);
+   * </ul>
+   *
+   * @return the root scope.
+   */
+  public static Scope openRootScope() {
+    synchronized (ROOT_SCOPES) {
+      if (ROOT_SCOPES.size() > 1) {
+        throw new RuntimeException(
+            "openRootScope() is not supported when multiple root scopes are enabled. Use 'Configuration.preventMultipleRootScopes()' to enable it.");
+      } else if (ROOT_SCOPES.size() == 1) {
+        return ROOT_SCOPES.values().iterator().next();
+      }
+      return openScope(Toothpick.class);
+    }
   }
 
   /**
@@ -127,27 +152,29 @@ public class Toothpick {
    * returned. Otherwise a new scope is created.
    *
    * @param name the name of the scope to open.
-   * @param shouldCheckMultipleRootScopes whether or not to check that the return scope is not
-   *     introducing a second tree in TP forest of scopes.
+   * @param isRootScope whether or not this is a root scope
    */
-  private static Scope openScope(Object name, boolean shouldCheckMultipleRootScopes) {
-    if (name == null) {
-      throw new IllegalArgumentException("null scope names are not allowed.");
-    }
+  private static Scope openScope(Object name, boolean isRootScope) {
+    synchronized (ROOT_SCOPES) {
+      if (name == null) {
+        throw new IllegalArgumentException("null scope names are not allowed.");
+      }
 
-    Scope scope = MAP_KEY_TO_SCOPE.get(name);
-    if (scope != null) {
+      Scope scope = MAP_KEY_TO_SCOPE.get(name);
+      if (scope != null) {
+        return scope;
+      }
+      scope = new ScopeImpl(name);
+      Scope previous = MAP_KEY_TO_SCOPE.putIfAbsent(name, scope);
+      if (previous != null) {
+        // if there was already a scope by this name, we return it
+        scope = previous;
+      } else if (isRootScope) {
+        ROOT_SCOPES.put(name, scope);
+        ConfigurationHolder.configuration.checkMultipleRootScopes(scope);
+      }
       return scope;
     }
-    scope = new ScopeImpl(name);
-    Scope previous = MAP_KEY_TO_SCOPE.putIfAbsent(name, scope);
-    if (previous != null) {
-      // if there was already a scope by this name, we return it
-      scope = previous;
-    } else if (shouldCheckMultipleRootScopes) {
-      ConfigurationHolder.configuration.checkMultipleRootScopes(scope);
-    }
-    return scope;
   }
 
   /**
@@ -157,17 +184,20 @@ public class Toothpick {
    * @param name the name of the scope to close.
    */
   public static void closeScope(Object name) {
-    // we remove the scope first, so that other threads don't see it, and see the next snapshot of
-    // the tree
-    ScopeNode scope = (ScopeNode) MAP_KEY_TO_SCOPE.remove(name);
-    if (scope != null) {
-      ScopeNode parentScope = scope.getParentScope();
-      if (parentScope != null) {
-        parentScope.removeChild(scope);
-      } else {
-        ConfigurationHolder.configuration.onScopeForestReset();
+    synchronized (ROOT_SCOPES) {
+      // we remove the scope first, so that other threads don't see it, and see the next snapshot of
+      // the tree
+      ScopeNode scope = (ScopeNode) MAP_KEY_TO_SCOPE.remove(name);
+      if (scope != null) {
+        ScopeNode parentScope = scope.getParentScope();
+        if (parentScope != null) {
+          parentScope.removeChild(scope);
+        } else {
+          ConfigurationHolder.configuration.onScopeForestReset();
+          ROOT_SCOPES.remove(name);
+        }
+        removeScopeAndChildrenFromMap(scope);
       }
-      removeScopeAndChildrenFromMap(scope);
     }
   }
 
--- a/toothpick-runtime/src/test/java/toothpick/ToothpickTest.java
+++ b/toothpick-runtime/src/test/java/toothpick/ToothpickTest.java
@@ -21,6 +21,7 @@ import static org.hamcrest.CoreMatchers.not;
 import static org.hamcrest.CoreMatchers.notNullValue;
 import static org.hamcrest.CoreMatchers.sameInstance;
 import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.equalToObject;
 import static org.junit.Assert.fail;
 import static org.mockito.Mockito.doNothing;
 import static org.mockito.Mockito.mock;
@@ -400,6 +401,44 @@ public class ToothpickTest {
 
     // THEN
     verify(scope, Mockito.atLeastOnce()).release();
+  }
+
+  @Test
+  public void openRootScope_shouldReturnTheRootScope_WhenRootScopeIsDefined() {
+    // GIVEN
+    Toothpick.openScopes("foo", "bar");
+
+    // WHEN
+    Scope rootScope = Toothpick.openRootScope();
+
+    // THEN
+    assertThat(rootScope, notNullValue());
+    assertThat(rootScope.getName(), equalToObject("foo"));
+  }
+
+  @Test
+  public void openRootScope_shouldReturnDefaultScope_WhenThereIsNoScopeTreeDefined() {
+    // GIVEN
+
+    // WHEN
+    Scope rootScope = Toothpick.openRootScope();
+
+    // THEN
+    assertThat(rootScope, notNullValue());
+    assertThat(rootScope.getName(), equalToObject(Toothpick.class));
+  }
+
+  @Test(expected = RuntimeException.class)
+  public void openRootScope_shouldThrowException_WhenThereAreMultipleScopeTreesDefined() {
+    // GIVEN
+    Toothpick.openScope("foo");
+    Toothpick.openScope("bar");
+
+    // WHEN
+    Toothpick.openRootScope();
+
+    // THEN
+    fail("Should throw an exception for multiple scope trees!");
   }
 
   @After
--- a/ktp/src/main/kotlin/toothpick/ktp/KTP.kt
+++ b/ktp/src/main/kotlin/toothpick/ktp/KTP.kt
@@ -41,6 +41,8 @@ object KTP {
         })
     }
 
+    fun openRootScope(): Scope = Toothpick.openRootScope()
+
     fun openScope(name: Any): Scope = Toothpick.openScope(name)
 
     fun openScope(name: Any, scopeConfig: Scope.ScopeConfig): Scope = Toothpick.openScope(name, scopeConfig)
--- a/ktp/src/test/kotlin/toothpick/ktp/KTPTest.kt
+++ b/ktp/src/test/kotlin/toothpick/ktp/KTPTest.kt
@@ -22,6 +22,7 @@ import org.amshove.kluent.shouldBeFalse
 import org.amshove.kluent.shouldBeTrue
 import org.amshove.kluent.shouldEqual
 import org.amshove.kluent.shouldNotBe
+import org.amshove.kluent.shouldNotBeNull
 import org.junit.jupiter.api.AfterEach
 import org.junit.jupiter.api.Test
 import toothpick.Scope
@@ -34,6 +35,17 @@ class KTPTest {
     @AfterEach
     fun tearDown() {
         Toothpick.reset()
+    }
+
+    @Test
+    fun `openRootScope should open default root scope`() {
+        // GIVEN
+
+        // WHEN
+        val scope = KTP.openRootScope()
+
+        // THEN
+        scope.shouldNotBeNull()
     }
 
     @Test


toothpick-runtime/src/main/java/toothpick/Toothpick.java
```diff
@@ -39,12 +39,31 @@
...
+   *     <ul>
+   *       <li>If there's an existing scope tree the root scope of that tree
+   *       <li>If there's no tree it will create a default root scope and return it
+   *       <li>If there are multiple scope trees (e.g. a forest) this method will throw an exception
+   *     </ul>
+   */
+  public static Scope openRootScope() {
+    if (ROOT_SCOPES.size() > 1) {
+      throw new RuntimeException(
+          "You have multiple root scopes defined. openRootScope() will work only with a single scope tree!");
```
dlemures: `openRootScope()` only works?

toothpick-runtime/src/main/java/toothpick/Toothpick.java
```diff
@@ -39,12 +39,31 @@
...
+   * @return The root scope as follows: <br>
+   *     <ul>
+   *       <li>If there's an existing scope tree the root scope of that tree
+   *       <li>If there's no tree it will create a default root scope and return it
+   *       <li>If there are multiple scope trees (e.g. a forest) this method will throw an exception
+   *     </ul>
+   */
+  public static Scope openRootScope() {
+    if (ROOT_SCOPES.size() > 1) {
+      throw new RuntimeException(
```
dlemures: what if we use `MultipleRootException`? The problem is that we have multiple roots. We can create a new constructor where we provide just a msg?

What do you think?

toothpick-runtime/src/main/java/toothpick/Toothpick.java
```diff
@@ -39,12 +39,31 @@
...
+          "You have multiple root scopes defined. openRootScope() will work only with a single scope tree!");
+    } else if (ROOT_SCOPES.size() == 1) {
+      return ROOT_SCOPES.values().iterator().next();
+    }
+    return openScope(Toothpick.class);
```
dlemures: I think that we could use a constant instead. That way if someone wanna add a child to the root scope the have 2 ways of using it:
1. With the new DSL: 
`openRootScope().openSubScope(this)`
2. But they can still use the one existing in TP 1.0 and 2.0, minimizing the changes on their code base:
`Toothpick.openScopes(Toothpick.ROOT_SCOPE, this)`

toothpick-runtime/src/main/java/toothpick/Toothpick.java
```diff
@@ -222,6 +242,7 @@ public static void inject(Object obj, Scope scope) {
    */
   private static void removeScopeAndChildrenFromMap(ScopeNode scope) {
     MAP_KEY_TO_SCOPE.remove(scope.getName());
+    ROOT_SCOPES.remove(scope.getName());
```
dlemures: This methods is meant to remove a subtree its head. So removing every element form the ROOT_SCOPES map does not make a lot of sense here.

I think it makes more sense to remove inside `closeScope`:
```java
public static void closeScope(Object name) {
    // we remove the scope first, so that other threads don't see it, and see the next snapshot of
    // the tree
    ScopeNode scope = (ScopeNode) MAP_KEY_TO_SCOPE.remove(name);
    if (scope != null) {
      ScopeNode parentScope = scope.getParentScope();
      if (parentScope != null) {
        parentScope.removeChild(scope);
      } else {
        ROOT_SCOPES.remove(name);    // <<<<<<< HERE <<<<<<<<<
        ConfigurationHolder.configuration.onScopeForestReset();
      }
      removeScopeAndChildrenFromMap(scope);
    }
  }
```

toothpick-runtime/src/main/java/toothpick/Toothpick.java
```diff
@@ -145,6 +164,7 @@ private static Scope openScope(Object name, boolean shouldCheckMultipleRootScope
       // if there was already a scope by this name, we return it
       scope = previous;
     } else if (shouldCheckMultipleRootScopes) {
```
dlemures: can we rename `shouldCheckMultipleRootScopes` to `isRootScope`?

`shouldCheckMultipleRootScopes` does not mean that the provided scope is a root scope. It is true that the value will be true if the scope is a root scope, so i think it is better to change the name.

toothpick-runtime/src/main/java/toothpick/Toothpick.java
```diff
@@ -39,12 +39,31 @@
...
     throw new RuntimeException("Constructor can't be invoked even via reflection.");
   }
 
+  /**
+   * @return The root scope as follows: <br>
+   *     <ul>
+   *       <li>If there's an existing scope tree the root scope of that tree
+   *       <li>If there's no tree it will create a default root scope and return it
+   *       <li>If there are multiple scope trees (e.g. a forest) this method will throw an exception
+   *     </ul>
```
dlemures: The javadoc does not match the style of the rest of the file. 

I'd have an explanation first, where you provide much of the details. Then the return value below with simple explanation to identify it. But the majority is in the details above.

toothpick-runtime/src/main/java/toothpick/Toothpick.java
```diff
@@ -39,12 +39,31 @@
...
+   */
+  public static Scope openRootScope() {
+    if (ROOT_SCOPES.size() > 1) {
```
dlemures: as this can have multiple threads at the same time, it would be better to save first `ROOT_SCOPES.size()` inside a variable and use it everywhere in the method. 

Otherwise it is possible that another thread will change the value in the middle.

toothpick-runtime/src/main/java/toothpick/Toothpick.java
stephanenicolas: I like the hidden constant that Mihaly is using and the fact the JVM ensure the unicity, as much as Android ensuresnthe unicity of the app scope.

toothpick-runtime/src/main/java/toothpick/Toothpick.java
```diff
@@ -39,12 +39,31 @@
   // its transformation
   private static final ConcurrentHashMap<Object, Scope> MAP_KEY_TO_SCOPE =
       new ConcurrentHashMap<>();
+  private static final ConcurrentHashMap<Object, Scope> ROOT_SCOPES = new ConcurrentHashMap<>();
```
stephanenicolas: Don't we need only a set ?

toothpick-runtime/src/main/java/toothpick/Toothpick.java
stephanenicolas: Use passive in error messages. And provide a solution. 
openRootScope is not supported when multiple root scopes are enabled. Use 'Configuration.xxx' to enable it.

toothpick-runtime/src/main/java/toothpick/Toothpick.java
stephanenicolas: Yes and ifnI remember correctly the check itself has some branches of its if logic that don't make any sense. This is highlighted in the code coverage of the class as well.

toothpick-runtime/src/test/java/toothpick/ToothpickTest.java
```diff
@@ -402,6 +404,44 @@ public void release_shouldCallScopeRelease() {
...
+  public void openRootScope_shouldReturnTheRootScope_WhenRootScopeIsDefined() {
+    // GIVEN
+    Toothpick.openScopes("foo", "bar");
+
+    // WHEN
+    Scope rootScope = Toothpick.openRootScope();
+
+    // THEN
+    assertThat(rootScope, notNullValue());
+    assertThat(rootScope.getName().toString(), is(equalTo("foo")));
```
stephanenicolas: I think you can remove to string and use 'same' matcher

toothpick-runtime/src/main/java/toothpick/Toothpick.java
dlemures: concurrenthashset does not exist

stephanenicolas: Ok, thx. Then add a comment in the code about the usage of a map in this case.

toothpick-runtime/src/main/java/toothpick/Toothpick.java
dlemures: So that means that if someone uses the root scope before a name is assigned to it, they will have to use only the new DSL?

toothpick-runtime/src/main/java/toothpick/Toothpick.java
```diff
@@ -39,12 +39,31 @@
...
 
+  /**
+   * @return The root scope as follows: <br>
+   *     <ul>
+   *       <li>If there's an existing scope tree the root scope of that tree
+   *       <li>If there's no tree it will create a default root scope and return it
+   *       <li>If there are multiple scope trees (e.g. a forest) this method will throw an exception
+   *     </ul>
+   */
+  public static Scope openRootScope() {
```
dlemures: this is not thread safe, we will have to synchnized this and also the place where we open the scopes (line 157):

```java
  private static Scope openScope(Object name, boolean shouldCheckMultipleRootScopes) {
    if (name == null) {
      throw new IllegalArgumentException("null scope names are not allowed.");
    }

    Scope scope = MAP_KEY_TO_SCOPE.get(name);
    if (scope != null) {
      return scope;
    }
    scope = new ScopeImpl(name);
    Scope previous = MAP_KEY_TO_SCOPE.putIfAbsent(name, scope); //  <----- HERE
    if (previous != null) {
      // if there was already a scope by this name, we return it
      scope = previous;
    } else if (shouldCheckMultipleRootScopes) {
      ConfigurationHolder.configuration.checkMultipleRootScopes(scope);
    }
    return scope;
  }
```

dlemures: Ideally try to synchronized if we have to open the scopes, not always. (similar to double-check for singletons)

dlemures:
I like the change!!! 

I added a few comments about the implementation.

toothpick-runtime/src/main/java/toothpick/Toothpick.java
code-twister: MultipleRootException has a defined message, and it's a RuntimeException, I'm not sure why we needed that in the first place...

dlemures: For the same reason there are lot of other subclasses of RuntimeExceptions, to make it more specific. Think for instance about `IllegalArgumentException`

toothpick-runtime/src/main/java/toothpick/Toothpick.java
code-twister: @dlemures can we do the synchronisation in a separate PR? that is a broader effort, not necessarily related to this change.

dlemures: The issue is that Toothpick is right now thread safe:
```java
 MAP_KEY_TO_SCOPE.putIfAbsent(name, scope); <--- This line alone is thread safe
```
But this change is breaking the safety. I think it would be nice not to merge it to master with concurrency working.

code-twister: so we should synchronise openRootScope and openScope? I'm not sure I understand what method of synchronisation you would like here...

toothpick-runtime/src/main/java/toothpick/Toothpick.java
```diff
@@ -39,12 +39,35 @@
...
+  // ConcurrentHashSet doesn't exist so we're not using a Set ;)
+  private static final ConcurrentHashMap<Object, Scope> ROOT_SCOPES = new ConcurrentHashMap<>();
   static Injector injector = new InjectorImpl();
 
   protected Toothpick() {
     throw new RuntimeException("Constructor can't be invoked even via reflection.");
   }
 
+  /**
+   * This function will return a root scope as follows:<br>
```
stephanenicolas: Remove 'this function as it is a method in oop and we avoid totological reference in doc. -> Opens the root scope.

toothpick-runtime/src/main/java/toothpick/Toothpick.java
```diff
@@ -39,12 +39,35 @@
...
+   *
+   * <ul>
+   *   <li>If there's an existing scope tree the root scope of that tree
+   *   <li>If there's no tree it will create a default root scope and return it
```
stephanenicolas: Remove upper case I, to all items and add a final ';' to all items please

toothpick-runtime/src/main/java/toothpick/Toothpick.java
```diff
@@ -39,12 +39,35 @@
...
+   *   <li>If there are multiple scope trees (e.g. a forest) this method will throw an exception
+   * </ul>
+   *
+   * @return the root scope
```
stephanenicolas: Add a final period.

toothpick-runtime/src/main/java/toothpick/Toothpick.java
stephanenicolas: We want to sync on the new map key set because size and put / get should be atomic. 
We should just merge when sync is done right.

toothpick-runtime/src/main/java/toothpick/Toothpick.java
stephanenicolas: Agree, it's a detail. Up to the OP IMHO.

toothpick-runtime/src/main/java/toothpick/Toothpick.java
stephanenicolas: I don't follow you @dlemures , this new API is totally optional and time independent.

toothpick-runtime/src/main/java/toothpick/Toothpick.java
stephanenicolas: Can we opt for a more fine grained lock here ?

toothpick-runtime/src/main/java/toothpick/Toothpick.java
```diff
@@ -179,7 +179,7 @@ private static Scope openScope(Object name, boolean isRootScope) {
    *
    * @param name the name of the scope to close.
    */
-  public static void closeScope(Object name) {
+  public static synchronized void closeScope(Object name) {
```
stephanenicolas: Same for both. Do we lock on the class elsewhere ?

toothpick-runtime/src/main/java/toothpick/Toothpick.java
code-twister: @dlemures can you tell me exactly the race condition that will cause an issue. I think we're ok with how openScope() already works.

toothpick-runtime/src/main/java/toothpick/Toothpick.java
dlemures: I mean that in order to open subscopes for the root scope we will need to use the new DSL api with the `openSubScope` method. It wont be possiblie to the same using the method `openScopes` @stephanenicolas

toothpick-runtime/src/main/java/toothpick/Toothpick.java
dlemures: Hey @code-twister sorry for the late reply, this week I will be off so I might not be able to reply.

The point is that whenever someone calls `openScope/s`(explicit root scope) from one thread together with `openRootScope` from another thread around the same time, there are 2 possibilities:

1. `openScope/s` is executed first and then `openRootScope` will open that same Scope.
2. `openRootScope` is executed first and then `openScope/s` will happen next and open a new scope (or crash if we have the configuration for it to happen).

That is a valid behavior (race condition) and TP is working as it should, it is the user responsibility to be aware of how `openRootScope` works (also we should make it very clear in the doc). So there is no issue there.

However, the current implementation there are situation where this method will not behave as expected:

**Example 1**
1. Initial situation: 0 root scopes
2. Thread 1:  finishes checking both `if statement` with false inside `openRootScope`.
3. Thread 2: opens new root scope with a custom name
4. Thread 1: opens a second root scope with name `Toothpick.class` and does not fail.

**Example 2**
1. Initial situation: 1 root scope
2. Thread 1: inside `} else if (ROOT_SCOPES.size() == 1) {` block
3. Thread 2: executes line `ROOT_SCOPES.remove(name);`
4. Thread 1: throws `NoSuchElementException` as the process was closed in the mean time

There might be more scenarios. My recommendation is to think about all possible combinations. It is possible that we might have something like:

**Valid race condition**
1. Initial situation: 1 root scope
2. Thread 1: inside `} else if (ROOT_SCOPES.size() == 1) {` block
3. Thread 2: opens a second root scope (check off)
4. Thread 1: will not fail and just return the first root scope from the map `ROOT_SCOPES`

But I consider that it is fine, as it is as if the user just opened the root scope after `openRootScope` is opened. And that doesnt happen the contract of the method/TP. However, the `example 1` above does as it is `openRootScope` the one opening the second scope.

Sorry for the long comment, i hope it is clear.

code-twister: It's clear... thanks @dlemures!
I will implement a solution soon.

dlemures: Actually, I think that **example 1** could be alright. It would be kind of equivalent to:

> openRootScope is executed first and then openScope/s will happen next and open a new scope (or crash if we have the configuration for it to happen).

code-twister:
coveralls doesn't seem to work properly...
getScope_shouldReturnAScope_whenThisScopeByThisKeyWasCreated should cover the line it reports being missed here: https://coveralls.io/builds/25500173/source?filename=toothpick-runtime%2Fsrc%2Fmain%2Fjava%2Ftoothpick%2FToothpick.java#L171

@stephanenicolas can you take a look, and let me know if I can do something to increase coverage

stephanenicolas:
@code-twister a test can't cover this line, or hardly.  It's a case of protection against multi-threading here. If the scope exists we return it, otherwise we create it and add it to the map. The line you mention would only be activated if a scope by the same name is added between [after the check to see if it existed by a different thread <--> before it is added to the map by this other thread]. 

We can't cover that by regular testing.

dlemures:
IT LOOKS AWESOME!!!

Thx @code-twister !

Releasing TP 3.1.0 with the feature!!!

PR #374, opened 2019-09-20
Fix dependencies in `smoothie-lifecycle`

### LINKED ISSUE #371: Dependencies of smoothie-lifecycle
Hello!
Why `smoothie-lifecycle` depends on `androidx.legacy:legacy-support-v4:1.0.0`? I've not found any usages of it.
And why it depends on `androidx.lifecycle:lifecycle-common-java8` but not on `androidx.lifecycle:lifecycle-common` (wihout java8)?
Same question about `smoothie-lifecycle-viewmodel` module.

osipxd:
- Removed `legacy-support-v4` from dependencies
- Removed unused dependencies from `smoothie-lifecycle`

Fixes #371


stephanenicolas:
@osipxd can you add some tests to make sure it works and there is no regression. Some tests for both java 7 & 8 & 9, what's needed to make sure we don't inadvertantly break it..

osipxd:
Looks like Gradle and Android Plugin does not support Java 7 and 9.
I figured out why build fails on my project with min-sdk 19 and doesn't on your example project with min-sdk 15. It is because you've set `targetCompatibility` and `sourceCompatibility` to `1.8` and I've not.

smoothie-lifecycle/build.gradle
```diff
@@ -30,11 +30,10 @@ android {
 dependencies {
   compileOnly project(':toothpick-runtime')
   api project(':smoothie')
-  implementation deps.inject
```
dlemures: we didn't need it?

smoothie-lifecycle-ktp/build.gradle
```diff
@@ -11,8 +11,8 @@ android {
   compileSdkVersion 28
 
   compileOptions {
-    sourceCompatibility JavaVersion.VERSION_1_8
-    targetCompatibility JavaVersion.VERSION_1_8
+    sourceCompatibility JavaVersion.VERSION_1_7
+    targetCompatibility JavaVersion.VERSION_1_7
```
dlemures: why?

stephanenicolas:
@osipxd so, do you keep the PR or your problem got solved ?

smoothie-lifecycle-ktp/build.gradle
osipxd: To make `smoothie-lifecycle` compatible with projects that use
```
sourceCompatibility JavaVersion.VERSION_1_7
targetCompatibility JavaVersion.VERSION_1_7
```
And because all other modules use 1.7 too

smoothie-lifecycle/build.gradle
osipxd: This module does not contain any JSR annotations

osipxd:
> @osipxd so, do you keep the PR or your problem got solved ?

The problem still actual for projects for the projects that use target jdk `1.7`.
Also, this PR removes unused dependencies.

stephanenicolas:
Thx @osipxd for this PR. However, I got some questions regarding it: 
* there are tests now, but do the tests fail before the PR and get green after the PR ? In other words, do the tests show a problem that is being solved by the PR ? Can the tests prevent a regression ?
* is there a real problem in the original code knowing that java 7 is obsolete ?

osipxd:
> * there are tests now, but do the tests fail before the PR and get green after the PR ? In other words, do the tests show a problem that is being solved by the PR ? Can the tests prevent a regression ?

I can try to make test fail before PR, and create a new branch that will demonstrate it (because this branch already contains fix).

> * is there a real problem in the original code knowing that java 7 is obsolete ?  

Unfortunately, Java 7 still used in many Android projects.

dlemures:
Besides the requirements about testing by @stephanenicolas , good on my side.

stephanenicolas:
@osipxd you can simply rebase and re-order your commits. But up to you if you feel more comfortable doing it another way.

Thanks for the clarification on the second point. I was not aware of the situation. 
Yes, definitely if you can show that there was a bug and you got a fix, and prevent regression, that would be awesome.

dimsuz:
From what I saw, many Android projects and libraries are moving to Java 8. Jake Wharton even said that community should *push* for updates by depending on new Java/SDK versions.

Current Android toolchain automatically desugars all Java 8 bytecode. At my work we are switching all Android projects to produce Java 8 bytecode: both Java and kotlin modules no problems at all so far.

That is to say, are you sure, you want to downgrade?

No big issue for me, as a user I will override this on a project level, but still...

osipxd:
@dimsuz, in my opinion, there is inconsistence that in the whole project used `sourceCompatibility '1.7'`, and only in `smoothie-lifecycle ` used `sourceCompatibility '1.8'`.  It should be the same for each module in the project. And if it is `1.8`, it is should be mentioned in the `Setup` section of the readme.

@stephanenicolas, now you can see that build (not tests) before the fix fails
I can make two pull requests. One for Java 7 support, and another one for removing unused dependencies.

stephanenicolas:
For the dependencies, @dlemures , can you have a look please ?

.travis.yml
```diff
@@ -39,7 +39,7 @@ cache:
 
 script:
   - ./gradlew clean check
-  - ./gradlew -b toothpick-sample/build.gradle clean check -x lint
+  - ./gradlew -b toothpick-sample/build.gradle clean build -x lint
```
dlemures: why this change?

dlemures: Is this to prove that the build fails for the sample project?

smoothie-androidx/build.gradle
```diff
@@ -28,9 +28,9 @@ dependencies {
   compileOnly project(':toothpick')
   api project(':smoothie')
   compileOnly deps.inject
-  compileOnly deps.androidxv4
+  compileOnly deps.androidxv7
 
-  testImplementation deps.androidxv4
+  testImplementation deps.androidxv7
```
dlemures: isn't `deps.fragment` enough on these 2 lines?

.travis.yml
osipxd: Yes, it is because `:build` fails on task `:mergeExtDexDebug`

smoothie-androidx/build.gradle
osipxd: Yes, it will be enough

stephanenicolas:
@osipxd after consulting our OSS devs and the other current contributors of TP, we agreed that we want to drop support for java 7 in TP3. We will document the change soon. 
Are you interested in closing this PR, and re-opening a PR with a matching ticket : 1 for the dependency change and one to make sure we enforce java 8 consistently in all modules ?

osipxd:
> @osipxd after consulting our OSS devs and the other current contributors of TP, we agreed that we want to drop support for java 7 in TP3. We will document the change soon.
> Are you interested in closing this PR, and re-opening a PR with a matching ticket : 1 for the dependency change and one to make sure we enforce java 8 consistently in all modules ?

I've created a new Issue for moving to Java 8. Also, I've removed commits related to Java 7 support.

dlemures:
Thx @osipxd :)

I'll release it probably next week once we have another PR ready.

Issue #370, opened 2019-09-17
Update ProGuard rules for Toothpick 3.0 and KTP


zawadz88:
Hi,
I'm migrating from TP 2.1.0 to TP 3.0 and I've noticed that some ProGuard rules need updating if you're using `@InjectConstructor` and KTP delegates instead of `@Inject`-annotated properties. I didn't find any info about this in the documentation so I'm posting it here.

The following seems to be working for me:
```
-keepnames @toothpick.InjectConstructor class *
-keepclasseswithmembernames class * { toothpick.ktp.delegate.* *; }
-keepclassmembers class * {
    toothpick.ktp.delegate.* *;
}
```
in addition to already defined rules. Note: I'm using that on top of the rules mentioned in #330

stephanenicolas:
Thx Piotr, we should update the wiki with this info.

stephanenicolas:
@zawadz88 can you make a PR on the proguard file that TP passes to downstream projects please ?  I have updated the docs already: https://github.com/stephanenicolas/toothpick/wiki/Proguard-&-TP

dimsuz:
By the way, are there any plans to put proguard rules inside the library jar's resources? Android's new shrinker R8 (now default) supports this, so even clients who do not use smoothie will receive them, no need for manual addition and supporting wiki page. Moshi, okhttp, retrofit use this feature, it's working great.

Maybe I'll create a PR one day, but currently rather busy, so at least keep this in mind, iirc this isn't a complex thing to do.

stephanenicolas:
Thx Piotr, take your time, as we wanna merge also another PR with this one
and release.

But thx for doing it,
 S.

stephanenicolas:
This has been addressed.

RumataEstorish:
When migrating to gradle plugin 7.0 had to add
-keep @javax.inject.Qualifier public class *
to keep qualifiers working

Issue #330, opened 2019-04-30
Smoothie Proguard config breaks Firebase


ultimate-deej:
I use Firebase Dynamic Links in my project. Turning ProGuard on caused `FirebaseDynamicLinks.getInstance()` to return `null`. After some digging it turned out that smoothie module was the cause.

At first removing this line seemed to help:
```-adaptclassstrings```
But I didn't want to remove smoothie from my project so I kept trying. After a while I came up with a rule that worked for me:
```-keep class * { @javax.inject.Qualifier <init>(...); }```

Btw, not being a ProGuard expert I would expect this to work instead:
```-keep @javax.inject.Qualifier class *```
But it doesn't 🤷‍♂️.

**Comment by stephanenicolas** [comment]
Can you show some code that you want to preserve, please?

Le mar. 30 avr. 2019 à 02:43, Maxim Naumov <notifications@github.com> a
écrit :

**Comment by ultimate-deej** [comment]
It's not my code I want to preserve. Instead, I want `FirebaseDynamicLinks.getInstance()` to return a non-null value. As to why that rule works or how it affects Firebase, no clue. I discovered it accidentaly, basically. But it suddenly works so I'll just stick with it until a better solution is found.

**Comment by stephanenicolas** [comment]
If I understand correctly,the qualifier annotation part is relevant to your
code.

If the option adaptclassstrings is impacting firebase, I would strongly
suggest that you open a ticket on their repo to ask them to document it.

TP might have required it for TP1 registries but I don't think it's needed
at all for TP2.

**Comment by ultimate-deej** [comment]
I couldn't find the Firebase Dynamic Links repo, this part of Firebase seems to be closed source. Anyway, I don't think it's a problem of Firebase. Because, isn't that rule too broad to be included in a library in the first place? And since it's not probably needed for TP2, could you consider removing it?

**Comment by ultimate-deej** [comment]
A clean project test showed that `-adaptclassstrings` alone doesn't break Firebase. I'll investigate the issue further.

**Comment by zawadz88** [comment]
I've encountered a similar issue when updating from TP 1.1.1 to 2.1.0.
The issue I'm having is with Firebase Performance though.
```
java.lang.NullPointerException: Attempt to invoke virtual method 'boolean com.google.firebase.perf.a.b()' on a null object reference
        at com.google.android.gms.internal.firebase-perf.fd.b(Unknown Source:2)
        at com.google.android.gms.internal.firebase-perf.fd.a(Unknown Source:0)
[...]

PR #376, opened 2019-09-25
Fixed Proguard rules for TP 3.x

zawadz88:
This is to fix issues #370 & #330 .

`-adaptclassstrings` was removed as it was breaking Firebase.

`toothpick.ktp.delegate` rules were added to work in Kotlin with `by inject()`, `by lazy()` etc. 

I've also removed `-keep @android.support.annotation.Keep class *` as it probably shouldn't be there in the first place (that's my best guess though, I might be missing something).

Moreover, I've added a sample rule on how to custom scope rules in your own projects (the one at the bottom).

I've tested this set of rules in our production app with R8 (ProGuard rules fallback version) & Firebase Performance plugin.

### DIFF
--- a/smoothie/proguard-rules.txt
+++ b/smoothie/proguard-rules.txt
@@ -2,18 +2,26 @@
 -keepattributes *Annotation*
 # Do not obfuscate classes with Injected Constructors
 -keepclasseswithmembernames class * { @javax.inject.Inject <init>(...); }
+-keepnames @toothpick.InjectConstructor class *
 # Do not obfuscate classes with Injected Fields
 -keepclasseswithmembernames class * { @javax.inject.Inject <fields>; }
 # Do not obfuscate classes with Injected Methods
 -keepclasseswithmembernames class * { @javax.inject.Inject <methods>; }
--keep @android.support.annotation.Keep class *
--dontwarn javax.inject.**
--dontwarn javax.annotation.**
+# Do not obfuscate classes with Inject delegates
+-keepclasseswithmembernames class * { toothpick.ktp.delegate.* *; }
+-keep class javax.inject.**
+-keep class javax.annotation.**
 -keep class **__Factory { *; }
 -keep class **__MemberInjector { *; }
-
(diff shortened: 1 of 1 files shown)


stephanenicolas:
It.looks good to me. I got 2 questions though: why do we need a rules for the delegates ? 
And aren't we missing one to keep the @scope annotated custom scope annotations ?

zawadz88:
Hi @stephanenicolas,
>why do we need a rules for the delegates?

We actually had crashes without this e.g.
```
Fatal Exception: toothpick.locators.NoFactoryFoundException: No factory could be found for class foo.bar.presentation.view.a. Check that the class has either a @Inject annotated constructor or contains @Inject annotated members.
       at toothpick.locators.FactoryLocator.getFactory + 37(FactoryLocator.java:37)
       at toothpick.ScopeImpl.lookupProvider + 359(ScopeImpl.java:359)
       at toothpick.ScopeImpl.getInstance + 93(ScopeImpl.java:93)
       at toothpick.ktp.delegate.EagerDelegate.onEntryPointInjected + 50(EagerDelegate.java:50)
       at toothpick.ktp.delegate.DelegateNotifier.notifyDelegates + 49(DelegateNotifier.java:49)
       at toothpick.ktp.KTP$1.inject + 37(KTP.java:37)
       at toothpick.Toothpick.inject + 213(Toothpick.java:213)
       at toothpick.ScopeImpl.inject + 140(ScopeImpl.java:140)
       ...

```

it seems it's the same reason why there are are rules for `@Inject` annotated properties and fields?

>aren't we missing one to keep the `@Scope` annotated custom scope annotations ?

I doesn't seem to be needed at least in our app. We have our own `@ActivityScope` annotation and the only rule we needed to add was:
```
-keepnames @foo.bar.ActivityScope class *
```
the annotation we use is in Java:
```
@javax.inject.Scope
@Documented
@Retention(RUNTIME)
public @interface ActivityScope {
}
```

dlemures:
great thx :)

PR #382, opened 2019-10-31
Fix RuntimeCheckOnConfiguration for android lower than API 19

MatnazaroFF:
`equals` method of `java.util.Objects` are available above API 19


dlemures:
Thx for the contribution!

PR #384, opened 2019-11-05
add scope config methods for both kotlin and java. Add a test. Might …

stephanenicolas:
…need more

We were missing a few overloads for the kotlin and java APIs of KTP and TP. 

We are missing tests for TP openRootScope / isRootScopeOpen.
We are missing tests that check if the lambda / scope config is applied correctly.

### DIFF
--- a/ktp/src/test/kotlin/toothpick/ktp/KTPTest.kt
+++ b/ktp/src/test/kotlin/toothpick/ktp/KTPTest.kt
@@ -40,12 +40,14 @@ class KTPTest {
     @Test
     fun `openRootScope should open default root scope`() {
         // GIVEN
+        KTP.isRootScopeOpen().shouldBeFalse()
 
         // WHEN
         val scope = KTP.openRootScope()
 
         // THEN
         scope.shouldNotBeNull()
+        KTP.isRootScopeOpen().shouldBeTrue()
     }
 
     @Test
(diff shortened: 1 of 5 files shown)


ktp/src/test/kotlin/toothpick/ktp/KTPTest.kt
```diff
@@ -68,7 +82,7 @@ class KTPTest {
         Toothpick.isScopeOpen("name").shouldBeFalse()
 
         // WHEN
-        val scope = KTP.openScope("name", Scope.ScopeConfig { })
+        val scope = KTP.openScope("name") { }
 
         // THEN
```
dlemures: test on Toothpick + more complete for both

ktp/src/test/kotlin/toothpick/ktp/KTPTest.kt
```diff
@@ -40,12 +40,26 @@ class KTPTest {
...
+    @Test
+    fun `openRootScope should open default root scope with a scope config`() {
+        // GIVEN
+
+        // WHEN
+        val scope = KTP.openRootScope {}
+
+        // THEN
+        scope.shouldNotBeNull()
+        KTP.isRootScopeOpen().shouldBeTrue()
```
dlemures: check if we applied the config

stephanenicolas:
@dlemures all comments are addressed. Your comments were not on lines that changed, that's why they don't appear updated.
Am I missing something (there was already a test for `TP.openScope(config)`).

Issue #420, opened 2020-09-26
The problem with cyclic injects checker


morder:
Hello!
Unfortunately, I got an unexpected crash with toothpick.

How to reproduce the crash

```
val parentScope = KTP.openScope("App")
val childScope = KTP.openScopes("App", "Child")

childScope.installModules(module {
   bind(...) // here is I bind some isolated(internal) instances that I need only in "Child" scope
   bind(IFace::class.java).to(Face::class.java).singleton() // also I bind some Interface to provide outside (like dagger's component)
})

// I want to provide some interface from my child scope to parent scope
parentScope.installModules(module {
    bind(IFace::class.java).toProviderInstance { childScope.getInstance(IFace::class.java) }.providesSingleton()
})
```

this code throws 
```
Class xxx.xxx.IFace creates a cycle:

                              ===============================
                             ||                             ||
                             \/                             ||
   xxx.xxx.IFace  ||
                             ||                             ||
                              ===============================
```

this code is working, but it creates instance immediately
```
     bind(IFace::class.java).toInstance(childScope.getInstance(IFace::class.java)).singleton()
```

If I remove any cycles checking from toothpick, everything is working as I expected

the problem with this code from `RuntimeCheckOnConfiguration.java`
```
final LinkedHashSet<Pair> linkedHashSet = cycleDetectionStack.get();
if (linkedHashSet.contains(pair)) { // pair = IFace|null
  throw new CyclicDependencyException(Pair.getClassList(linkedHashSet), clazz);
}
```
**we are trying to get `IFace` instance from `child` scope, but we already trying to get it from parent scope**
cycles checking code can't understand it :)

afaucogney:
Hi, this usecase is a bit strange to me, could you please explain the real usecase ? What do you need to achieve ?Why not to just inject Iface in the parentscope as well as in the child scope ? If the binding is done in the parent, the child scope will benefit from this binding if not overriden.Additionally, a the scope structure is a tree, it allow by design to have second child scope. How would you solve this ?Everyscope have to be sufficient, so a rootscope has to live/survive without any child scope from my point of view (at least).Sent from Android deviceLe 26 sept. 2020 09:57, morder <notifications@github.com> a écrit :
Hello!
Unfortunately, I got an unexpected crash with toothpick.
How to reproduce the crash
val parentScope = KTP.openScope("App")
val childScope = KTP.openScopes("App", "Child")

childScope.installModules(module {
[...]

PR #428, opened 2021-06-17
Fixed dependency cycle checker

morder:


morder:
fixing issue
https://github.com/stephanenicolas/toothpick/issues/420

Issue #429, opened 2021-06-18
Easy interface binding


morder:
Dagger is allowing to do smthg like this
```
interface IFooDi : IFoo, IFooUi

@Binds
@AppScope
abstract fun fooDi(instance: Foo): IFooDi

@Binds
@AppScope
abstract fun foo(instance: IFooDi): IFoo

@Binds
@AppScope
abstract fun fooUi(instance: IFooDi): IFooUi
```

but with toothpick, you should write a lot of provider classes like 

```
bind(IFooDi::class.java).to(Foo::class.java).singleton()
bind(IFooUi::class.java).toProvider(FooUiProvider::class.java).providesSingleton()
bind(IFoo::class.java).toProvider(FooProvider::class.java).providesSingleton()

class FooUiProvider @Inject constructor(
    private val instance: IFooDi
) : Provider<IFooUi> {
    override fun get() = instance
}

class FooProvider @Inject constructor(
    private val instance: IFooDi
) : Provider<IFoo> {
    override fun get() = instance
}
```

PR #430, opened 2021-06-18
cyclic interface binding

morder:


morder:
issue fixing 
https://github.com/stephanenicolas/toothpick/issues/429

PR #433, opened 2022-01-19
Setup Gtihub Actions CI (opened by mistake, sorry)

outadoc:


afaucogney:
@outadoc We call this a spoil ! But that's a good one ! If you need help, feel free to ask.

